PR #16, opened 2023-03-30
Add ArrayTypeWrapper to create a StringArray wrapper class

Maddin-619:
I added a ArrayTypeWrapper class equivalent to the existing scalar TypeWrapper for handling c-arrays safely.

The new StringArray class is a specific ArrayTypeWrapper with additional functions to convert between vectors of strings and c-array UA_Strings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 # generate compile_commands.json
-set(CMAKE_EXPORT_COMPILE_COMMANDS ON) 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 # compiled binaries folders (same as open62541)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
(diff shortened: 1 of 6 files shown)


lukasberbuer:
Thanks for the pull request @Maddin-619.
What is your use-case for this wrapper class?

I only used e.g. `std::vector<std::string>` in combination with `opcua::Variant` to write arrays to a variable node. The `opcua::Variant` already has methods to convert container of `std::string`s to arrays of `UA_String`. Depending on the memory layout you may use:
- for native types, e.g. `UA_String`:
  - `opcua::Variant::setArray(T* array, size_t size)`
  - `opcua::Variant::setArray(std::vector<T>& array)`
- for types convertible to native types:
  - `opcua::Variant::setArrayCopy(InputIt first, InputIt last)`
  - `opcua::Variant::setArrayCopy(const T* array, size_t size)`
  - `opcua::Variant::setArrayCopy(const std::vector<T>& array)`

More info in the [opcua::Variant documentation](https://open62541pp.github.io/open62541pp/classopcua_1_1Variant.html).

Maddin-619:
Oh man. I dind't realised that variants are capable of wrapping arrays.
I want to use the [UA_CreateCertificate](https://github.com/open62541/open62541/blob/73e7f41a44c979afe6b15741b50ebc964fdb544e/plugins/include/open62541/plugin/create_certificate.h#L35-L40) function of the open62541 library which takes to UA_String arrays. I testet it with your suggestion and it worked. But actualy the deref inside the getArray function have to be removed:
https://github.com/open62541pp/open62541pp/blob/79fbe961b6f800f74670b0becdd5ba40435a7b2f/include/open62541pp/types/Variant.h#L243

lukasberbuer:
Glad this works for you!
Just fixed the dereference bug in #25 - thanks for finding 👍

PR #17, opened 2023-03-30
Update open62541 to v1.3.5

Maddin-619:
I updated open62541 to the latest stable version 1.3.5. There where only a few breaking changes which I fixed.


lukasberbuer:
Thanks Martin! I wasn't aware of those breaking changes.
I merged it into a feature branch for now, because we need to add some `ifdef`s to keep compatibility with the older open62541 versions.

Btw, all CI and compatibility pipelines should be enabled for pull requests - will change that soon :)
I will have a look at your other pull requests on Monday 👍

PR #18, opened 2023-03-30
Feature/read write ua data value

Maddin-619:
There was a commented out writeDataValue method inside the Node. I implemented these method alongside a new DataValue wrapper type. It's now possible to read and write node values with timestamp and status code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,6 +100,7 @@ endif()
 # open62541pp library (static/shared depending on option BUILD_SHARED_LIBS)
 add_library(
     open62541pp
+    src/DataValue.cpp
     src/Helper.cpp
     src/Node.cpp
     src/NodeId.cpp
--- a/include/open62541pp/DataValue.h
+++ b/include/open62541pp/DataValue.h
@@ -1,0 +1,22 @@
+#pragma once
+
+#include "open62541pp/TypeWrapper.h"
+#include "open62541pp/Variant.h"
+
+namespace opcua {
+
+class DataValue : public TypeWrapper<UA_DataValue, UA_TYPES_DATAVALUE> {
+public:
+    using TypeWrapperBase::TypeWrapperBase;
+
+    void setVariant(const Variant& variant);
+    Variant getVariant() const;
+
+    void setSourceTimeStamp(const DateTime& ts);
+    DateTime getSourceTimeStamp() const;
+
+    void setStatusCode(const UA_StatusCode);
+    UA_StatusCode getStatusCode() const;
+};
+
+}  // namespace opcua
--- a/include/open62541pp/Node.h
+++ b/include/open62541pp/Node.h
@@ -5,6 +5,7 @@
 #include <string_view>
 #include <vector>
 
+#include "open62541pp/DataValue.h"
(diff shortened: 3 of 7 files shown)


lukasberbuer:
Hi @Maddin-619,
I already was working on this feature locally so I had to do a little bit of merging now. I was missing the setter methods. Those are added in commit [dd2b2f6](https://github.com/open62541pp/open62541pp/commit/dd2b2f6ceebde8ecec47ab51fdb2c995682e89a1).

I propose to use `std::optional` as the return type for the getter methods to make handling with the optional fields more explicit.

I like the convenience of following `Node::writeScalar` helpers:

```cpp
/// Write scalar with timestamp to variable node.
template <typename T, Type type = detail::guessType<T>()>
void writeScalar(const T& value, const DateTime& datetime);

/// Write scalar with timestamp and status code to variable node.
template <typename T, Type type = detail::guessType<T>()>
void writeScalar(const T& value, const DateTime& datetime, UA_StatusCode code);
```

I'm not sure how this approach will scale in the future. My first issue is, that the `datetime` can be either `serverTimestamp` or `sourceTimestamp`. We can not add overloads for both versions because of the same type.
On the other hand, we also should add `Node::writeArray` with the same interfaces which results in a lot of convenience methods.

Instead we could add helper functions `createDataValueScalar` and `createDataValueArray` and pass the return values to the `Node::writeDataValue` method. Those functions could have following signature:

```cpp
template <typename T, Type type = detail::guessType<T>()>
DataValue createDataValueScalar(
    const T& value,
    std::optional<DateTime> sourceTimestamp = {},
    std::optional<DateTime> serverTimestamp = {},
    std::optional<UA_StatusCode> statusCode = {}
);

const auto dv1 = createDataValueScalar(11.11f, DateTime::now());
const auto dv2 = createDataValueScalar(11.11f, DateTime::now(), DateTime::now());
const auto dv3 = createDataValueScalar(11.11f, DateTime::now(), {}, UA_STATUSCODE_BAD);

// and maybe another one with full options (source and server picoseconds)
```

What do you think of that solution?

Btw., sorry for the merge conflicts because of my late pushs to remote. I really appreciate your contributions!

lukasberbuer:
Hi @Maddin-619,

I tried to come up with a solution. First, I like the `Node::writeScalar` / `Node::writeArray` methods, because they choose the most efficient way to . On the other hand, I don't like the logic to be part of the `Node` class.
I created static methods `Variant::fromScalar` and `Variant::fromArray` which will create variants the most efficient way. `Node` is using those logics now.

Additionally, I added a rvalue overload of `DataValue::setValue`. Now we can create `DataValue`s from scalars or arrays with no copies if possible:

```cpp
float number = 11.11f;

DataValue dv;
dv.setValue(Variant::fromScalar(number));  // no copy here, Variant and DataValue both reference `number`
dv.setSourceTimestamp(DateTime::now());

Node.writeDataValue(dv);
```

It's a little bit more code than your solution, but we don't have to create more `Node` methods.
Does this work for you?

Maddin-619:
This works great! Your api is acctualy much more flexible than my suggestion with the overloaded method, even if it is a little more code. Thank you very much!

Issue #20, opened 2023-04-05
Plans for the Client-implementation?


ErikTempelaarVO:
Hi,

I'm planning to start development using open62541 and I came across your library looking for a neat C++ wrapper.
I want to build a client and that is on your TODO-list.
Do you have any plans for it already? Some code in a draft-branch perhaps? Are you open to contributions?

lukasberbuer:
Hi @ErikTempelaarVO,

thanks for bringing that up. I made some drafts but didn't come up with I solution I was happy with. The main reason was, that I want to stick with the `Node` class. The `Node` class itself carries the `NodeId` and a `std::shared_ptr` to the underlying `Server` connection (to maybe extend the lifetime of the connection).
We need to introduce a `Client` class for sure. How can we adapt the `Node` class to handle both server and client connections? I was thinking of following solutions:
1. `NodeBase` base class with `NodeServer` and `NodeClient` implementations. Virtual functions will add some overhead. And we loose the beauty of returning `Node`s by value.
2. Use a `std::variant` to store either a server or client connection in the `Node`. Then we need dispatching in every `Node` method, keep the same interface but have no (?) performance penalties.
3. Introduce a `NodeClient` with the same or similar interface.

I saw that @duca started working on a client on [a fork](https://github.com/duca/open62541pp). It seems like he did choose option 3. Maybe we can share ideas and collaborate on this feature. I'll try to invite him to this discussion :)

lukasberbuer:
Draft of how option 2 might look like: [Link to godbolt](https://godbolt.org/z/78P7T54rM)
I like the approach with the free functions for both server and client - which might be useful for some use-cases.

```cpp
#include <iostream>
#include <utility>
#include <variant>

template<class... Ts> struct overload : Ts... { using Ts::operator()...; };
template<class... Ts> overload(Ts...) -> overload<Ts...>;

class Server {};
class Client {};
class NodeId {};
class Variant {};


// free functions for both Server and Client in separate namespace
namespace services {

void writeValue(Server& server, const NodeId& id, const Variant& var) {
    std::cout << "Server writeValue\n";
}

void writeValue(Client& client, const NodeId& id, const Variant& var) {
    std::cout << "Client writeValue\n";
}

}  // namespace services


class Node {
public:
    Node(std::variant<Server, Client> connection)
        : connection_(std::move(connection)) {}

    void writeValue(const Variant& var) {
        // std::visit(overload{
[...]

PR #22, opened 2023-04-07
feat: free functions in services namespace

lukasberbuer:
As discussed in #20 to prepare client implementation

### DIFF
--- a/include/open62541pp/Node.h
+++ b/include/open62541pp/Node.h
@@ -8,7 +8,8 @@
 #include "open62541pp/Common.h"
 #include "open62541pp/Server.h"
 #include "open62541pp/TypeConverter.h"  // guessType
-#include "open62541pp/TypeWrapper.h"
+#include "open62541pp/services/Attribute.h"
+#include "open62541pp/services/NodeManagement.h"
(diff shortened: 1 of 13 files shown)


include/open62541pp/Node.h
duca, line 1: One thing that I might have forgotten. Will we keep both Node and NodeClient or when the client is ready for prime time we are going to adjust Node to be able to use either Client or Server ?

tests/Node.cpp
```diff
@@ -16,85 +13,6 @@ TEST_CASE("Node") {
...
-
-        SECTION("Unspecified dimension (ValueRank <= 0)") {
-            auto valueRank = GENERATE(
-                ValueRank::Any,
-                ValueRank::Scalar,
-                ValueRank::ScalarOrOneDimension,
-                ValueRank::OneOrMoreDimensions
-            );
-            CAPTURE(valueRank);
-            node.writeValueRank(valueRank);
```
duca, line 66: These tests could be migrated to the unit testing of the Attribute functions

duca (commented):
Left some comments, but except for identifying an opportunity to create tests for the Attribute package functions, I see no changes to be made

PR #23, opened 2023-04-11
client pieces

duca:
This has the prototype NodeClient and Client to be used as starting point of the Client implementation. Currently Client is severely lacking if compared with the Server class, however NodeClient is very similar to Node in functionality. Some features were commented out just because I didn't have the time to finalize and test them.

I am new to PRs on Github, so pardon me for opening one to master instead of a separate client-specific branch

PR #33, opened 2023-04-14
feat: basic client implementation

lukasberbuer:
Prior discussions in #20 and #23.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,6 +98,7 @@ endif()
 add_library(
     open62541pp
     src/Client.cpp
+    src/CustomLogger.cpp
     src/Helper.cpp
     src/Server.cpp
     src/services/Attribute.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,7 @@ endif()
 # open62541pp library (static/shared depending on option BUILD_SHARED_LIBS)
 add_library(
     open62541pp
+    src/Client.cpp
     src/Helper.cpp
     src/Server.cpp
     src/services/Attribute.cpp
--- a/doc/Doxyfile.in
+++ b/doc/Doxyfile.in
@@ -52,6 +52,9 @@ HTML_OUTPUT            = "${doxygen_html_dir}"
 HTML_COLORSTYLE        = LIGHT
 EXT_LINKS_IN_WINDOW    = YES
 
+# Preprocessor
+MACRO_EXPANSION        = YES
+
 # Dot diagrams
 HAVE_DOT               = YES
 DOT_FONTNAME           = Helvetica
--- a/doc/DoxygenLayout.xml
+++ b/doc/DoxygenLayout.xml
@@ -4,6 +4,7 @@
     <tab type="mainpage" visible="yes" title="Overview"/>
 
     <tab type="user" url="@ref opcua::Server" title="Server"/>
+    <tab type="user" url="@ref opcua::Client" title="Client"/>
     <tab type="user" url="@ref opcua::Node" title="Node"/>
     <tab type="user" url="@ref TypeWrapper" title="Wrapper"/>
     <tab type="user" url="@ref Services" title="Services"/>
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,4 +1,5 @@
 set(examples_sources
+    client_connect.cpp
     client_find_servers.cpp
     server.cpp
     server_instantiation.cpp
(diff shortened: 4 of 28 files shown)


lukasberbuer:
@duca, I just focused on the discovery so far. There are some decision I'm unsure about:
1. `Client::findServers` and `Client::getEndpoints` could be static methods. Disadvantage: The client config can not be modified, for example to adjust timeouts. Or we add an argument to pass an optional `UA_ClientConfig`.
2. `Client` could be constructed with the `serverUrl` directly so we don't have to pass it to the `connect` method etc. This makes primarily sense if the methods of 1 are static. The Python library `asyncua` chooses this approach: https://opcua-asyncio.readthedocs.io/en/latest/api/asyncua.client.html#asyncua.client.client.Client

duca:
> @duca, I just focused on the discovery so far. There are some decision I'm unsure about:
> 
> 1. `Client::findServers` and `Client::getEndpoints` could be static methods. Disadvantage: The client config can not be modified, for example to adjust timeouts. Or we add an argument to pass an optional `UA_ClientConfig`.
> 2. `Client` could be constructed with the `serverUrl` directly so we don't have to pass it to the `connect` method etc. This makes primarily sense if the methods of 1 are static. The Python library `asyncua` chooses this approach: https://opcua-asyncio.readthedocs.io/en/latest/api/asyncua.client.html#asyncua.client.client.Client

Regarding 1., I think it is fine to pass a UA_ClientConfig (or a wrapper for it) as argument to the static methods

Regarding 2: agreed, matches fine with 1 approach and allows for simple RAII-esque approach

lukasberbuer:
@duca, I had some time and implemented the `NodeManagement` and `Attribute` service set for the client. The free functions made those additions really easy. I'm quite happy with the outcome.

One of the most important client features is missing: browsing! But this is also missing for the server so far.
Maybe we can merge this branch first and work on the browsing feature for both.

Do you have any further suggestions, API proposals etc.?

lukasberbuer:
> > @duca, I just focused on the discovery so far. There are some decision I'm unsure about:
> > 
> > 1. `Client::findServers` and `Client::getEndpoints` could be static methods. Disadvantage: The client config can not be modified, for example to adjust timeouts. Or we add an argument to pass an optional `UA_ClientConfig`.
> > 2. `Client` could be constructed with the `serverUrl` directly so we don't have to pass it to the `connect` method etc. This makes primarily sense if the methods of 1 are static. The Python library `asyncua` chooses this approach: https://opcua-asyncio.readthedocs.io/en/latest/api/asyncua.client.html#asyncua.client.client.Client
> 
> Regarding 1., I think it is fine to pass a UA_ClientConfig (or a wrapper for it) as argument to the static methods
> 
> Regarding 2: agreed, matches fine with 1 approach and allows for simple RAII-esque approach

I fear that we will introduce too much overhead with the static methods. I just saw that the types `UA_ClientConfig` and `UA_ServerConfig`  are huge (> 500 bytes). The configs have to be created with every `findServer` or `getEndpoints` call if those are static methods...

duca:
> > > @duca, I just focused on the discovery so far. There are some decision I'm unsure about:
> > > 
> > > 1. `Client::findServers` and `Client::getEndpoints` could be static methods. Disadvantage: The client config can not be modified, for example to adjust timeouts. Or we add an argument to pass an optional `UA_ClientConfig`.
> > > 2. `Client` could be constructed with the `serverUrl` directly so we don't have to pass it to the `connect` method etc. This makes primarily sense if the methods of 1 are static. The Python library `asyncua` chooses this approach: https://opcua-asyncio.readthedocs.io/en/latest/api/asyncua.client.html#asyncua.client.client.Client
> > 
> > 
> > Regarding 1., I think it is fine to pass a UA_ClientConfig (or a wrapper for it) as argument to the static methods
> > Regarding 2: agreed, matches fine with 1 approach and allows for simple RAII-esque approach
> 
> I fear that we will introduce too much overhead with the static methods. I just saw that the types `UA_ClientConfig` and `UA_ServerConfig` are huge (> 500 bytes). The configs have to be created with every `findServer` or `getEndpoints` call if those are static methods...

What if we had Client and Server as a global or as a singleton? We would need to set them up once and forget, right?

duca:
> @duca, I had some time and implemented the `NodeManagement` and `Attribute` service set for the client. The free functions made those additions really easy. I'm quite happy with the outcome.
> 
> One of the most important client features is missing: browsing! But this is also missing for the server so far. Maybe we can merge this branch first and work on the browsing feature for both.
> 
> Do you have any further suggestions, API proposals etc.?

I do have a working browse implementation on the client side, the API is extremely simple:

https://github.com/duca/open62541pp/blob/master/include/open62541pp/Browser.h
https://github.com/duca/open62541pp/blob/master/src/Browser.cpp

duca:
I added a constructor in my NodeClient class that accepted a RefereceDescription besides the client and it is usable right now.

PR #39, opened 2023-04-22
added composed types ReferenceDescription and ApplicationDescription

duca:
Using the asWrapper api. Changed the CMakeLists.txt to reflect that change and adjusted the enum size in Common.h to adhere to the proper type size.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,21 +105,28 @@ add_library(
     src/services/Attribute.cpp
     src/services/Browse.cpp
     src/services/NodeManagement.cpp
+    src/types/ApplicationDescription.cpp
     src/types/Builtin.cpp
     src/types/DataValue.cpp
     src/types/DateTime.cpp
     src/types/NodeId.cpp
+    src/types/ReferenceDescription.cpp
     src/types/Variant.cpp
 )
 add_library(open62541pp::open62541pp ALIAS open62541pp)
-target_include_directories(open62541pp PUBLIC include)
+target_include_directories(open62541pp INTERFACE
+        $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+        PUBLIC
+        $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+        )
 target_link_libraries(
     open62541pp
     PUBLIC
         open62541::open62541
     PRIVATE
         Threads::Threads
-        open62541pp_project_options
+        $<BUILD_INTERFACE:open62541pp_project_options>
 )
 
 # tests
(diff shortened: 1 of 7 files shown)


lukasberbuer:
Thanks @duca. Sadly, the wrappers for `ReferenceDescription` and `ApplicationDescription ` are already in the master in `types/Composed.h`. I was quite busy on this project last week.

Following snippet is interesting:
```cpp
    bool const& isForward() const noexcept;
    bool& isForward() noexcept;
```

1. You return primitives also as references so we don't need setters anymore. I think that's smart. I was thinking about `ReferenceDescription` as a return type which is never really created, so I just returned `bool` from the `getIsForward()` method.
2. You don't prefix with `get` in this specific case. I was also struggling with this decision. I don't like the `get`s so be everywhere in the interface. On the other hand, sometimes setters make sense and then we get a naming problem. Still unsure about that.

lukasberbuer:
PS: I got very lazy in #40 and implemented `ReferenceDescription` with the help of macros like this:

```cpp
class ReferenceDescription
    : public TypeWrapper<UA_ReferenceDescription, UA_TYPES_REFERENCEDESCRIPTION> {
public:
    using TypeWrapperBase::TypeWrapperBase;

    UAPP_COMPOSED_GETTER_WRAPPER(NodeId, getReferenceTypeId, referenceTypeId)
    UAPP_COMPOSED_GETTER(bool, getIsForward, isForward)
    UAPP_COMPOSED_GETTER_WRAPPER(ExpandedNodeId, getNodeId, nodeId)
    UAPP_COMPOSED_GETTER_WRAPPER(QualifiedName, getBrowseName, browseName)
    UAPP_COMPOSED_GETTER_WRAPPER(LocalizedText, getDisplayName, displayName)
    UAPP_COMPOSED_GETTER_CAST(NodeClass, getNodeClass, nodeClass)
    UAPP_COMPOSED_GETTER_WRAPPER(ExpandedNodeId, getTypeDefinition, typeDefinition)
};
```

duca:
> Thanks @duca. Sadly, the wrappers for `ReferenceDescription` and `ApplicationDescription ` are already in the master in `types/Composed.h`. I was quite busy on this project last week.
> 
> Following snippet is interesting:
> 
> ```c++
>     bool const& isForward() const noexcept;
>     bool& isForward() noexcept;
> ```
> 
> 1. You return also primitives as references to avoid setters. I think that's smart. I was thinking about `ReferenceDescription` as a return type which is never really created, so I just return by value from the `getIsForward()` method. Returning by value for this kind of return type should be a little bit more performant but we loose the possibility to assign.
> 2. You don't prefix with `get` in this specific case. I was also struggling with this decision. I don't like the `get`s so be everywhere in the interface. On the other hand, sometimes setters make sense and then we get a naming problem. Still unsure about that.

No problem, Lukas, last week I had to travel and got very little time to work on these projects. I will close the PR, then

duca:
No problem, Lukas, last week I had to travel and got very little time to work on these projects. I will close the PR, then

PR #40, opened 2023-04-22
feat: browse/view service set

lukasberbuer:
Implement overloaded functions in `services` namespace of [View service set](https://reference.opcfoundation.org/Core/Part4/v105/docs/5.8):
- `browse`
- `browseNext`
- `translateBrowsePathToNodeIds`
- `browseSimplifiedBrowsePath`


duca (commented):
Lately I have been thinking if using std::array was better in the long run considering that we usually receive simultaneously the size and the pointer to the data.

lukasberbuer:
> Lately I have been thinking if using std::array would be better in the long run considering that we usually receive simultaneously the size and the pointer to the data.

`std::span` would be perfect in my opinion. Because of the poiner-intercovertibility between native types and wrapper types, we could return a span of wrappers without any copies. But this requires C++20. Another option would be `gsl::span` or some other implementations, like this: https://github.com/tcbrindle/span

PR #45, opened 2023-05-05
feat: subscription & monitored item service set

lukasberbuer:
- Implementation of the Subscription and MonitoredItem service set as free functions in namespace `services`.
- Add high-level `Subscription<T>` and `MonitoredItem<T>` classes


lukasberbuer:
@duca, let's move the discussions to this place :)

PR #53, opened 2023-05-27
fix: incorrect monitoreditem ids

### LINKED ISSUE #52: MonitoredItemIds are not correct
When subscribing to multiple nodes the MonitoredItemIDs do not change.
This results in incorrect NodeIDs/AttributeIDs, when fetching from inside the callback.

`Data change notification:
- subscription id:   21
- monitored item id: 1
- node id:           i=2258
- attribute id:      13
Current server time (UTC): 09:28:22
Data change notification:
- subscription id:   21
- monitored item id: 1
- node id:           i=2258
- attribute id:      13
The Answer: 215`

The mixup happens here:

`
static const MonitoredItem<Client> monitoredItem(*connectionPtr, subId, monId);
callback(monitoredItem, value);
`

The resulting MonitoredItem<Client> is always monId '1'. Even if the passed monId = 2.
Removing the static helps.

image

ErikTempelaarVO:
This fixes #52

### DIFF
--- a/src/Subscription.cpp
+++ b/src/Subscription.cpp
@@ -138,7 +138,7 @@ MonitoredItem<Client> Subscription<Client>::subscribeDataChange(
         [connectionPtr = &connection_, callback = std::move(onDataChange)](
             uint32_t subId, uint32_t monId, const DataValue& value
         ) {
-            static const MonitoredItem<Client> monitoredItem(*connectionPtr, subId, monId);
+            const MonitoredItem<Client> monitoredItem(*connectionPtr, subId, monId);
             callback(monitoredItem, value);
         }
     );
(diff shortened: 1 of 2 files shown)


ErikTempelaarVO:
I looked into increasing the coverage to include events.
But that's not trivial considering the events are not imlemented on the server, and not easy to trigger in a testcase. Right?

It would be nice to test against some sort of mock, and not the actual server.

ErikTempelaarVO:
I think the existing checks could be improved. It now only checks inequality, but a comparison to the id that is returned via the call to the subscrption would be better.

lukasberbuer:
This looks good, thank you! I was travelling the last days but will have a detailled look tomorrow and merge it.

Regarding the tests/mocks: I am not an expert in C++ testing and mocking. I hesitated first to do end to end tests with servers and clients but then  saw that open62541 is doing the same. Do you have a better solution to test e.g. client functionalities?

Issue #59, opened 2023-07-02
Access level control at client


johanneskopf:
Hi,

I want to:
1.  create a **parent node** at the server from the client.
2. create an **int node** at the server from the client as a child node of **parent node** and set the attributes accordingly.

Here is a sample code:
```cpp
#include "open62541pp/open62541pp.h"

int main() {
    opcua::Client client;
    client.connect("opc.tcp://localhost:4840");
    auto objectsNode = client.getObjectsNode();
    auto parentNode = objectsNode.addObject({1, "parent.node"}, "parent node");
    auto integerNode = parentNode.addVariable({1, "int.node"}, "int node");
    integerNode.writeDataType(opcua::Type::Int32); // this line throws the error below
    integerNode.writeDisplayName({"en-US", "int node"});
    integerNode.writeDescription({"en-US", "int node"});
    integerNode.writeScalar<int32_t>(42);
}
```

This code always throws the following error:
> terminate called after throwing an instance of 'opcua::BadStatus'
> what():  BadUserAccessDenied

Due to the descriptive error, I figured I need to set the correct access level or write mask at the integer node to be able to write the attributes after creating the node at the server.

So I tried to add the following code in-between, but that resulted in the same error being thrown:
```cpp
    integerNode.writeAccessLevel(UA_ACCESSLEVELMASK_READ | UA_ACCESSLEVELMASK_WRITE);
    integerNode.writeUserAccessLevel(UA_ACCESSLEVELMASK_READ | UA_ACCESSLEVELMASK_WRITE);
    integerNode.writeWriteMask(~0U);
    integerNode.writeUserWriteMask(~0U);
```

How to achieve this?

lukasberbuer:
Hi @johanneskopf,
I can reproduce the issue but have to dig deeper.. 🕳️
Hopefully I can provide you a solution soon.

lukasberbuer:
A similar issue was also reported for open62541: https://github.com/open62541/open62541/issues/3910
The mentioned status code `2147614720` = `0x80200000` = `BadIdentityTokenInvalid` differs but is the adjacent one:
```
BadUserAccessDenied,0x801F0000,User does not have permission to perform the requested operation.
BadIdentityTokenInvalid,0x80200000,The user identity token is not valid.
```

lukasberbuer:
I could narrow down the problem: It makes a difference if the attributes are defined during construction of the node or afterwards. The `writeMask`'s default value is `0` (no permission):
[...]

PR #68, opened 2023-07-23
feat: add variable with attributes

johanneskopf:
My take on adding a variable and set some attributes such as accessLevel , write mask, datatype, and initial value.

### DIFF
--- a/include/open62541pp/Node.h
+++ b/include/open62541pp/Node.h
@@ -91,6 +91,22 @@ public:
         return {connection_, id, false};
     }
 
+    /// @copydoc services::addVariable
+    template<typename T, Type type = detail::guessType<T>()>
+    Node addVariable(
+            const NodeId& id,
+            std::string_view browseName,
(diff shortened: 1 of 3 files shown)


lukasberbuer:
Hi @johanneskopf, I like your minimalistic solution to this problem. On the other hand I was thinking about different use case scenarios why you might need initial node attributes.
One scenario came into my mind where this approach doesn't work well: For example if a client wants to create a read-only variable node with a static value and data type. Following steps are necessary:
1. Client creates a variable node with e.g. `writeMask = ~0U`
2. Client writes Value and DataType attribute to node
3. Client sets write mask to read-only again: `writeMask = 0U`

By not allowing to set all attributes initially we make certain use cases way more complex. What's your take on this?

johanneskopf:
I like the fluent interface approach we were discussing in https://github.com/open62541pp/open62541pp/issues/59, thus I am closing this PR.

PR #72, opened 2023-07-23
chore: fix cmake code in readme

johanneskopf:
Fix cmake code in the main readme which does not compile as is.


lukasberbuer:
Thanks @johanneskopf 👍

Issue #71, opened 2023-07-23
Callbacks missing on server: onConnected/onDisconnected/onSessionActivated/onSessionClosed


johanneskopf:
Client side has the following callbacks implemented:
```cpp
void Client::onConnected(StateCallback callback) {
    setStateCallback(getContext(), ClientState::Connected, std::move(callback));
}

void Client::onDisconnected(StateCallback callback) {
    setStateCallback(getContext(), ClientState::Disconnected, std::move(callback));
}

void Client::onSessionActivated(StateCallback callback) {
    setStateCallback(getContext(), ClientState::SessionActivated, std::move(callback));
}

void Client::onSessionClosed(StateCallback callback) {
    setStateCallback(getContext(), ClientState::SessionClosed, std::move(callback));
}
```

Now I need those callbacks on the server side too. Is it possible to implement them? I looked at the wrapper code and found that a similar implementation as on the client side is not possible, as the client side has a state-callback (ua_client.c):
```cpp
if(client->config.stateCallback)
  client->config.stateCallback(client, client->channel.state, client->sessionState, client->connectStatus);
```
whereas the server side hasn't. To use those callbacks on the server-side, it is instead necessary to override the callbacks in the access control of the configuration:
```cpp
config->accessControl.activateSession = ...;
config->accessControl.closeSession = ...;
```
It would be nice if these callbacks were also available on the server side of the wrapper.

johanneskopf:
Given this also is part of the server-side access-control, it is related to https://github.com/open62541pp/open62541pp/issues/69.

lukasberbuer:
Hi @johanneskopf. This is a tricky one. We could hook into the `UA_AccessControl` callbacks `activateSession` and `closeSession`, but those are usually used to provide access and clean-up session-related storage. We have to wrap the `UA_AccessControl` to call both the state callback and the actual access callback. But we can not prevent users from using the C-API and providing their own access control and un-hooking our wrapper.
Both `onConnected` and `onDisconnected` seem to be impossible - I don't see any spot where to hook in.
May I ask what you need those state callbacks for?

johanneskopf:
Thank you for taking a look @lukasberbuer . Basically, I need a counter on the server side that always tells how many clients are currently connected and still active. But just the activateSession and closeSession callbacks are not enough. I created an issue two months ago hoping to get more information/ideas in the open62541 repository:  https://github.com/open62541/open62541/issues/5788

lukasberbuer:
[...]

Issue #69, opened 2023-07-23
Server access control callbacks


johanneskopf:
Hi,
I need to be able to control the access to the server nodes, methods, etc. Hereby, open62541 provides many callbacks in the UA_AccessControl class. To achieve my goal, I need to overwrite them. Optimally, the wrapper would provide some functionality to do so.

For now, I tried to use the open62541 directly to achieve this, but those callbacks all have a `UA_Server *server` as argument. E.g.
```cpp
UA_UInt32 (*getUserRightsMask)(UA_Server *server, UA_AccessControl *ac, const UA_NodeId *sessionId, void *sessionContext, const UA_NodeId *nodeId, void *nodeContext);
```
How can I convert the `UA_Server *server` argument to an `opcua::Server` object? Getting the `UA_Server *server` from an `opcua::Server` object is possible by using the method `UA_Server* handle() noexcept;`, but I need it the other way round. That way, I could use the opcua::Server functionality inside the callback.

Example code how to override the callbacks:
```cpp
#define UA_ACCESSLEVELMASK_ALL_BYTE 0xFF
#define UA_ACCESSLEVELMASK_ALL_UINT32 0xFFFFFFFF

static UA_UInt32 getUserRightsMaskOwn(UA_Server *server, UA_AccessControl *ac, const UA_NodeId *sessionId, void *sessionContext, const UA_NodeId *nodeId, void *nodeContext) {
    printf("getUserRightsMaskOwn called\n");
    return UA_ACCESSLEVELMASK_ALL_UINT32;
}

static UA_Byte getUserAccessLevelOwn(UA_Server *server, UA_AccessControl *ac, const UA_NodeId *sessionId, void *sessionContext, const UA_NodeId *nodeId, void *nodeContext) {
    printf("getUserAccessLevelOwn called\n");
    return UA_ACCESSLEVELMASK_ALL_BYTE;
}

static UA_Boolean getUserExecutableOwn(UA_Server *server, UA_AccessControl *ac, const UA_NodeId *sessionId, void *sessionContext, const UA_NodeId *methodId, void *methodContext) {
    printf("getUserExecutableOwn called\n");
    return true;
}

static UA_Boolean getUserExecutableOnObjectOwn(UA_Server *server, UA_AccessControl *ac, const UA_NodeId *sessionId, void *sessionContext, const UA_NodeId *methodId, void *methodContext, const UA_NodeId *objectId, void *objectContext) {
[...]

PR #74, opened 2023-07-25
feat: access control

lukasberbuer:
Wrapper of `UA_AccessControl` plugin with `Server` integration. As requested in #69 and #71.

Design idea:
- `AccessControlBase` base class with pure virtual methods
- `AccessControlDefault` class with same logic as `UA_AccessControl_default`
- Custom access control logics can be implemented by deriving from either the base class or `AccessControlDefault` and overriding (some) methods
- Hook `AccessControlBase&` into `Server`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,7 @@ endif()
 include(GNUInstallDirs)
 add_library(
     open62541pp
+    src/AccessControl.cpp
     src/Client.cpp
     src/Crypto.cpp
     src/CustomLogger.cpp
--- a/include/open62541pp/AccessControl.h
+++ b/include/open62541pp/AccessControl.h
@@ -1,0 +1,287 @@
+#pragma once
+
+#include <any>
+#include <vector>
+
+#include "open62541pp/Auth.h"  // Login
+#include "open62541pp/Config.h"
+#include "open62541pp/types/Builtin.h"
+#include "open62541pp/types/Composed.h"
+#include "open62541pp/types/ExtensionObject.h"
+
+// forward declare
+struct UA_AccessControl;
+
+namespace opcua {
+
+// forward declare
+class Server;
+
+/**
+ * Access control base class.
+ *
+ * Used to authenticate sessions and grant access rights accordingly.
+ * Custom access control can be implemented by deriving from this class and overwriting the access
+ * control callbacks.
+ *
+ * The `sessionId` and `sessionContext` can originally be both `NULL` in open62541.
+ * This is the case when, for example, a MonitoredItem (the underlying Subscription) is detached
+ * from its Session but continues to run.
+ * This wrapper passes `sessionId` and `sessionContext` by const reference, so they can't be `NULL`.
+ * Instead, an empty `NodeId` and `AccessControlBase::SessionContext` will be created.
+ *
+ * @see UA_AccessControl
+ * @see https://www.open62541.org/doc/1.3/plugin_accesscontrol.html
+ */
+class AccessControlBase {
+public:
+    /// Arbitrary data attached to a session.
+    using SessionContext = std::any;
+
+    explicit AccessControlBase(Server& server);
+
+    virtual ~AccessControlBase();
+
+    AccessControlBase(const AccessControlBase& other) = delete;
+    AccessControlBase(AccessControlBase&& other) noexcept = delete;
+
+    AccessControlBase& operator=(const AccessControlBase& other) = delete;
+    AccessControlBase& operator=(AccessControlBase&& other) noexcept = delete;
+
+    /// Get available user token policies.
+    virtual std::vector<UserTokenPolicy> getUserTokenPolicies() noexcept = 0;
+
+    /**
+     * Authenticate a session.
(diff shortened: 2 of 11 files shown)


johanneskopf:
Hi @lukasberbuer ,
Works great so far, didn't encounter any issue so far. Since I can now override the `activateSession` and `closeSession` methods myself, that is most likely all I need for my use case. Thanks! :)

Just some features I would like to see added/questions that came up:
1) `hash()` method needed to create a set of `NodeId`, e.g.:
`std::unordered_set<NodeId>`
The needed code for this to work would be:
```
namespace std {
    template<>
    struct hash<opcua::NodeId> {
        std::size_t operator()(const opcua::NodeId& id) const noexcept {
            return id.hash();
        }
    };
}
```
2) What is the easiest way given a NodeId to find out whether a node exists? I came up with the following solution, but I am interested if there is an easier/cleaner solution here.
```
bool doesNodeExist(Session &session, const NodeId &nodeId) {
    bool nodeExists = true;
    try {
        auto _ = session.getConnection().getNode(nodeId); // checks for existence!
    }
    catch (const BadStatus &status) {
        if (status.code() == UA_STATUSCODE_BADNODEIDUNKNOWN)
            nodeExists = false;
        else
            throw status;
    }
    return nodeExists;
}
```
3) A method which, given a `NodeId X` and a `NodeId Y`, finds out whether `Y` is reachable starting from `X` by either traversing Forward or Inverse. My idea was to implement this by going over all nodes in queue `Q`, starting with `X`. Thereby, when visiting a node I first check whether I already visited the node (using set `S`). Second, I add all references of this node to `Q`  (using `nodeId.browseReferencedNodes(BrowseDirection::Inverse);`). Third, I add the node itself to `S`, to prevent visiting this node twice.
Is there a built in method that accomplishes this as well / is this something that would be useful to add or do you think this use case is too specific?

lukasberbuer:
Hi @johanneskopf, thanks for testing :)
Internally, I'm already using `onSessionActivated` and `onSessionClosed` callbacks to register the active sessions. I wouldn't be complicated to make them public. Just let me know if this has any benefit.

Regarding your questions:

1. Good point. We could easily add this to open62541pp. I guess it makes sense?

2. From the client side, this seems to be the only sensible way. Internally, I'm just reading the `NodeId` attribute to check if the node exists. From the server side, there are more efficient mechanisms using the `UA_Nodestore` (`nodestore` member of `UA_ServerConfig`):

```c
// nodestore.h

typedef struct {
    /* Nodestore context and lifecycle */
    void *context;
    void (*clear)(void *nsCtx);
    // ...
    /* ``Get`` returns a pointer to an immutable node. ``Release`` indicates
     * that the pointer is no longer accessed afterwards. */
    const UA_Node * (*getNode)(void *nsCtx, const UA_NodeId *nodeId);

    void (*releaseNode)(void *nsCtx, const UA_Node *node);
    // ...
} UA_Nodestore;
```

The check could be like this:

```cpp
auto* config = UA_Server_getConfig(server.handle());
const auto nodeId = server.getObjectsNode().getNodeId();
// const auto nodeId = opcua::NodeId(1, 1000);
const auto* node = config->nodestore.getNode(config->nodestore.context, nodeId.handle());
const bool exists = (node != nullptr);
std::cout << "NodeId " << nodeId.toString() << " exists: " << exists << std::endl;
if (node != nullptr) {
    config->nodestore.releaseNode(config->nodestore.context, node);
}
```

What would be a good place to add such functions?
- `bool Client/Server::isNodeId(const NodeId&)`
- `bool services::exists(ServerOrClient&, const NodeId&)`
- `static bool Node::exists(const NodeId&)`
- ...?

3. Sounds like you are searching for the `UA_Server_browseRecursive` function. This already checks that nodes are not visited twice. Should be easy to wrap this in open62541pp. Shall we do it?

```c
// server.h

/* Non-standard version of the Browse service that recurses into child nodes.
 *
 * Possible loops (that can occur for non-hierarchical references) are handled
 * internally. Every node is added at most once to the results array.
 *
 * Nodes are only added if they match the NodeClassMask in the
 * BrowseDescription. However, child nodes are still recursed into if the
 * NodeClass does not match. So it is possible, for example, to get all
 * VariableNodes below a certain ObjectNode, with additional objects in the
 * hierarchy below. */
UA_StatusCode UA_EXPORT UA_THREADSAFE
UA_Server_browseRecursive(UA_Server *server, const UA_BrowseDescription *bd,
                          size_t *resultsSize, UA_ExpandedNodeId **results);
```

johanneskopf:
Hi @lukasberbuer ,
1. I believe it makes sense. If anyone has the need to add an `unordered_set` or `unordered_map` of NodeId's, it will definitely be appreciated.
2. I also vote for your current favorite `bool NodeId::exists(Server&)` / `bool NodeId::exists(Client&)`.
3. That's exactly what I was looking for. Already tried it and that method gets the job done so far for me (still more testing required though). So yes, please wrap that method! :)

lukasberbuer:
Thanks for the feedback @johanneskopf. I will target the discussed points in separate PRs and merge this branch now.

Issue #70, opened 2023-07-23
Add custom datatype


johanneskopf:
I need to add a custom data type to the server, and then use this custom data type to add a node of that type. It should be possible to read/write to/from this node from the server and client side. I have created an example (only server side) that is implemented using the open62541 library directly. Now I want to create an improved version using this wrapper, but I could not find out how to achieve a similar functionality. Is this possible?

```cpp
#include <open62541/server.h>
#include <open62541/server_config_default.h>

#include <csignal>

char localeStr[] = "en-US";

static volatile UA_Boolean running = true;

static void stopHandler(int sig) {
    running = false;
}

static UA_NodeId generateNodeIdWithString(UA_UInt16 namespaceIndex, char *string) {
    UA_String uaString = {sizeof(string) - 1, reinterpret_cast<UA_Byte *>(string)};
    UA_NodeId node;
    node.namespaceIndex = namespaceIndex;
    node.identifierType = UA_NODEIDTYPE_STRING;
    node.identifier.string = uaString;
    return node;
}

#define ID_DATA_TYPE_TWO_STRINGS 4400
#define BINARY_ENCODING_ID_TWO_STRINGS  1
const UA_NodeId TwoStringsVariableTypeId = {1, UA_NODEIDTYPE_NUMERIC, {4401}};

static char twoStringsNodeIdStr[] = "TwoStrings";
UA_NodeId twoStringsNodeId = generateNodeIdWithString(1, twoStringsNodeIdStr);

typedef struct {
    UA_String a;
    UA_String b;
} TwoStrings;

static UA_DataTypeMember TwoStrings_members[2] = {
        {UA_TYPENAME("a") &UA_TYPES[UA_TYPES_STRING], 0, false, false},
        {UA_TYPENAME("b") &UA_TYPES[UA_TYPES_STRING], 0, false, false}
};

static const UA_DataType TwoStringsType = {
        UA_TYPENAME("TwoStrings") {1, UA_NODEIDTYPE_NUMERIC, {ID_DATA_TYPE_TWO_STRINGS}}, {1, UA_NODEIDTYPE_NUMERIC, {BINARY_ENCODING_ID_TWO_STRINGS}}, sizeof(TwoStrings), UA_DATATYPEKIND_STRUCTURE,
        false, false, 2, TwoStrings_members
};

static void addDataTypeTwoStrings(UA_Server *server) {
    UA_DataTypeAttributes attr = UA_DataTypeAttributes_default;
[...]

PR #76, opened 2023-07-29
feat: custom datatype

lukasberbuer:
As discussed in #70.

**Goal**: Integrate custom types and easy and safe as possible.

Instead of specifying the `UA_DataType` with its members of `UA_DataTypeMember` manually, the `DataTypeBuilder` can be used to create type definitions and to derive the type attributes from the type itself. Most checks are executed at compile time.
Especially the offset calculation of members is tricky. There are some experimental `constexpr` solutions discussed here: https://gist.github.com/graphitemaster/494f21190bb2c63c5516
If we can implement a `constexpr` version of `offsetOf`, we can move more checks to compile time.

### Example: How to create a struct data type

```cpp
struct S {
    int16_t a;
    float* b;
    float* c;
};

const auto dt = DataTypeBuilder<S>::createStructure("S", {1, 1001} /* typeId */, {1, 1} /* binaryEncodingId */)
    .addField<&S::a>("a")
    .addField<&S::b>("b")
    .addField<&S::c>("c")
    .build();
```

### Example: Structs with wrapper types

Is it safe to use the `UA_DataType` definition of `T` for the wrapper type `TypeWrapper<T>`?

According to the standard, `TypeWrapper<T>` and `T` are pointer-interconvertible, because:

> Two objects a and b are pointer-interconvertible if: one is a [standard-layout](https://en.cppreference.com/w/cpp/language/data_members#Standard_layout) class object and the other is the first non-static data member of that object or any base class subobject of that object
> -- [cppreference](https://en.cppreference.com/w/cpp/language/static_cast#pointer-interconvertible)

So this guarantees that there is no padding **before** the first and only member of the `TypeWrapper<T>`.
BUT it is not guaranteed, that `sizeof(T) == sizeof(TypeWrapper<T>)` although this is usually the case.
So there might be additional padding **after** the wrapper type.

Some interesting reads:
- https://stackoverflow.com/questions/66597260/does-c-guarantee-identical-binary-layout-for-trivial-structs-with-a-single-t
- https://stackoverflow.com/questions/55650798/guaranteed-memory-layout-for-standard-layout-struct-with-a-single-array-member-o

Using wrapper classes inside custom types is not a problem. The size of the wrapper and the padding between the struct members are derived by the `DataTypeBuilder` or can be easily computed manually with `sizeof` and `offsetof`.

```cpp
struct SWrapper {
    opcua::Guid guid;
    int value;
};

auto dt = DataTypeBuilder<SWrapper>::createStructure("SWrapper", {1, 1005}, {1, 5})
    .addField<&SWrapper::guid>("guid")
    .addField<&SWrapper::value>("value")
    .build();
```

### DIFF
--- a/src/DataType.cpp
+++ b/src/DataType.cpp
@@ -247,7 +247,7 @@ bool operator==(const UA_DataType& lhs, const UA_DataType& rhs) noexcept {
     }
 #ifdef UA_ENABLE_TYPEDESCRIPTION
     if (std::strcmp(emptyIfNullptr(lhs.typeName), emptyIfNullptr(rhs.typeName)) != 0) {
-        return false;
+        return false;  // NOLINT
     }
 #endif
     return true;
--- a/src/DataType.cpp
+++ b/src/DataType.cpp
@@ -186,11 +186,6 @@ void DataType::setMembers(const std::vector<DataTypeMember>& members) {
 }
 
 bool operator==(const UA_DataTypeMember& lhs, const UA_DataTypeMember& rhs) noexcept {
-#ifdef UA_ENABLE_TYPEDESCRIPTION
-    if (std::strcmp(emptyIfNullptr(lhs.memberName), emptyIfNullptr(rhs.memberName)) != 0) {
-        return false;
-    }
-#endif
 #if UAPP_OPEN62541_VER_GE(1, 3)
     if (lhs.memberType != rhs.memberType) {
         return false;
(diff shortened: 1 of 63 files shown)


src/DataType.cpp
```diff
@@ -0,0 +1,328 @@
...
+    }
+    return *this;
+}
+
+// NOLINTNEXTLINE
+const char* DataType::getTypeName() const noexcept {
+#ifdef UA_ENABLE_TYPEDESCRIPTION
+    return handle()->typeName;
+#endif
+    return nullptr;
```
johanneskopf, line 90: Fix compile warning by changing this to:
```
#ifdef UA_ENABLE_TYPEDESCRIPTION
    return handle()->typeName;
#else
    return nullptr;
#endif
```

lukasberbuer: Thanks @johanneskopf, just fixed it.

PR #85, opened 2023-08-11
feat!: ArrayView

lukasberbuer:
View a contiguous sequence of objects, similar to `std::span` in C++20.

ArrayViews are used to return open62541 arrays without copy and to use them with the standard library algorithms. The view just holds two members: the pointer to `T` and the size, so it's lightweight and trivially copyable.


lukasberbuer:
Renamed branch to `feat/span`

PR #91, opened 2023-08-15
feat: services::browseRecursive

### LINKED ISSUE #89: Wrap `UA_Server_browseRecursive`
Challenge: No corresponding implementation for `UA_Client`.

As discussed in #74.

lukasberbuer:
Closes #89.

### DIFF
--- a/include/open62541pp/services/View.h
+++ b/include/open62541pp/services/View.h
@@ -10,6 +10,7 @@
 namespace opcua {
 class ByteString;
 class QualifiedName;
+class Server;
 }  // namespace opcua
 
 namespace opcua::services {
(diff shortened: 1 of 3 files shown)


lukasberbuer:
@johanneskopf, I was thinking about adding a method to the `Node` class. But because there is no `Client` implementation, I would prefer to only have the free function and keep the symmetry of the `Node` class. Any remarks from your side?

johanneskopf:
@lukasberbuer  I see. By "free function" you mean that the method is in the `services` namespace, right? That also works great for me and given the circumstances is a good solution. I have already update my code with the new wrapper method and it works perfectly fine👍

lukasberbuer:
> @lukasberbuer I see. By "free function" you mean that the method is in the `services` namespace, right? That also works great for me and given the circumstances is a good solution. I have already update my code with the new wrapper method and it works perfectly fine👍

Yes exactly. In the end, the `Node` class is just a more convenient way of using the free functions in the `services` namespace.

Thanks for testing!

PR #92, opened 2023-08-15
feat!: check if node id exists

### LINKED ISSUE #88: Check if `NodeId` exists
Add new methods for efficient checks if a `NodeId` actually exists.

As discussed in #74.

lukasberbuer:
Closes #88.

Initial ideas was to add following methods:
- `bool NodeId::exists(Server&)`
- `bool NodeId::exists(Client&)`

This would introduce client and server logic to the `NodeId` data type, which is by definition agnostic to the actual server-client-connection. Instead, the `Node` class seems to be a better place, because it couples `Server`/`Client` and the actual `NodeId`.

Following changes are proposed:
- Remove the optional existance check in the `Node` constructor with the `checkExists` flag
- Add a method `Node::exists` to explicitly check the existance of a node

If just the existance of a `NodeId` should be checked, a `Node` object must be created. BUT the construction is very cheap.

@johanneskopf: Any remarks/ideas?

### DIFF
--- a/include/open62541pp/Node.h
+++ b/include/open62541pp/Node.h
@@ -34,14 +34,9 @@ template <typename ServerOrClient>
 class Node {
 public:
     /// Create a Node object.
-    /// @exception BadStatus (BadNodeIdUnknown) If `checkExists` enabled and `id` not found
-    Node(ServerOrClient& connection, NodeId id, bool checkExists = true)
+    Node(ServerOrClient& connection, NodeId id)
         : connection_(connection),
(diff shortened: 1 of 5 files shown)


lukasberbuer:
@johanneskopf, please review!
I figured out, that the server-side test can even be simpler:

```cpp
template <>
bool Node<Server>::exists() noexcept {
    void* context{};
    const auto status = UA_Server_getNodeContext(getConnection().handle(), getNodeId(), &context);
    return (status == UA_STATUSCODE_GOOD);
}
```

PR #108, opened 2023-09-14
feat: callAsync function

lukasberbuer:
Async client version of calling a server method.


lukasberbuer:
Continued in PR #111.

PR #110, opened 2023-09-17
build: gcc 7 support

lukasberbuer:
Workaround to allow GCC 7 with partial C++ 17 support: https://en.cppreference.com/w/cpp/compiler_support/17

Use `<experimental/filesystem>` instead of `<filesystem>`.

### DIFF
--- a/include/open62541pp/types/Builtin.h
+++ b/include/open62541pp/types/Builtin.h
@@ -2,11 +2,20 @@
 
 #include <array>
 #include <cstdint>
-#include <filesystem>
 #include <iosfwd>  // forward declare ostream
 #include <string>
 #include <string_view>
 #include <vector>
+
+// Workaround for GCC 7 with partial C++17 support
+// https://github.com/open62541pp/open62541pp/issues/109
+#if !__has_include(<filesystem>) && __has_include(<experimental/filesystem>)
+#include <experimental/filesystem>
+namespace fs = std::experimental::filesystem;
+#else
+#include <filesystem>
+namespace fs = std::filesystem;
+#endif
 
 #include "open62541pp/ErrorHandling.h"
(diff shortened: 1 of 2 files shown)


piperoc:
Thank you for this. I did build and test it on GCC 7.1, 7.4, 11.0 on both x86 and my original ARM targets (in fact both my Altera SOCFPGA and a Raspberry PI which are armhf).
Everything tests and runs as expected.
Let me know if you need more information.

lukasberbuer:
Thanks for testing @piperoc! Let's merge it.

ErikTempelaarVO:
I appreciate the effort but unfortunately my GCC 7 doesn't have these includes.
I'm trying to compile on GCC 7.2 (from msys2) and 7.5 (mxe) and they both fail.
I have other includes in experimental, but not filesystem.h

Haven't figured out what is different for my toolchains, but it looks like another valid reason to upgrade to a newer GCC. :)

lukasberbuer:
@ErikTempelaarVO, did you try to compile some minimal example like this (with `-std=c++17` flag)?

```cpp
#include <experimental/filesystem>

int main() {
    std::experimental::filesystem::path path{};
}
```

ErikTempelaarVO:
Thanks for the suggestion, I have c++17 enabled already.
I don't think this can be easily solved, but maybe this can help anyone else.

I think that for these older versions of GCC the flag '--enable-libstdcxx-filesystem-ts' is required.
Or at least it was required for some time for the mingw-variants of GCC.

See:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78870#c14
https://github.com/mxe/mxe/issues/2220
https://gcc.gnu.org/legacy-ml/libstdc++/2019-11/msg00120.html

PR #111, opened 2023-09-18
feat: async services

lukasberbuer:
Implement async alternatives to free functions in `services` namespace (client only).
The async functions have the suffix `*Async` and share the same function signature.

This PR adapts the well-proven [asynchronous model of (Boost) Asio](https://think-async.com/asio/asio-1.28.0/doc/asio/overview/model.html). A key goal of Asio's asynchronous model is to support multiple composition mechanisms. This is achieved via completion tokens.

## Completion tokens

Completion tokens are accepted as the final argument of asynchronous operations.

```cpp
// Function signature of the completion handler: void(opcua::StatusCode, opcua::Variant&)
template <typename CompletionToken = opcua::DefaultCompletionToken>
auto opcua::services::readValueAsync(
    opcua::Client& client,
    const opcua::NodeId& id,
    CompletionToken&& token = opcua::DefaultCompletionToken()
);
```

Following completion tokens can be used:

- A callback function (object) with the expected completion signature:
  ```cpp
  void(opcua::StatusCode code)              // for void result types
  void(opcua::StatusCode code, T& result);  // for non-void result types
  void(opcua::StatusCode code, T result);   // for non-void result types (trivially-copyable)
  ```
- `opcua::useFuture` to return a future object `std::future<T>` (**default**)
- `opcua::useDeferred` to return a callable for deferred execution
- `opcua::useDetached` to detach the asynchronous operation
- Custom tokens by providing template specializations for `opcua::AsyncResult`


include/open62541pp/services/detail/ClientService.h
```diff
@@ -0,0 +1,114 @@
...
+                Response& response = *static_cast<Response*>(responsePtr);
+                auto result = opcua::detail::tryInvoke(
+                    std::get<TransformResponse>(*context), response
+                );
+                if constexpr (std::is_void_v<Result>) {
+                    std::invoke(handler, result.code());
+                } else {
+                    std::invoke(handler, result.code(), *result);
+                }
+            } catch (const std::exception&) {
```
chandryan, line 63: This catch would currently swallow uncaught std exceptions from the completion handler, for example a std::out_of_range etc., right?

Wouldn't it be better to let the exception be uncaught and let the application potentially crash, in case the completion handler doesn't take care of it?
If there is a scenario where such exceptions need to be caught outside of the completion handler, a user could invoke client.run() within a try-catch block.

doc/async_model.md
```diff
@@ -0,0 +1,118 @@
...
+    client,
+    id,
+    [](opcua::StatusCode code, opcua::Variant& result) {
+        // ...
+    }
+);
+```
+
+The status code `opcua::StatusCode` needs to be checked to determine the validity of the result.
+The callback is executed within the client's or server's event loop. Please make sure not to block the event loop with synchronous requests.
```
chandryan, line 50: It looks like you have gained some experience already with this case, could you explain what is happening in this scenario?

Looking at the code in open62541 I thought it should work to run the event loop in both threads, the one using the sync API and the one running only the event loop. Though it would not be clear in which thread the completion handler is executed.

include/open62541pp/detail/Result.h
```diff
@@ -0,0 +1,207 @@
...
+private:
+    StatusCode code_;
+};
+
+/**
+ * Result encapsulates a result value and a status code.
+ * A result may contain just an error status code, just the return value or both.
+ */
+template <typename T>
+class Result {
```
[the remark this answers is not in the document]
lukasberbuer: I was thinking about that as well. Do you think the interface of the `Result` class is suitable to make it public?
Using the `useFuture` token should still result in `std::future<T>` and not `std::future<Result<T>>`, do you agree?

doc/async_model.md
```diff
@@ -0,0 +1,118 @@
...
+Open62541++ adapts the the well-proven [asynchronous model of (Boost) Asio](https://think-async.com/asio/asio-1.28.0/doc/asio/overview/model.html). A key goal of Asio's asynchronous model is to support multiple composition mechanisms. This is achieved via completion tokens.
+
+## Completion tokens
+
+Open62541++ accepts completion tokens as the final argument of asynchronous operations.
+
+```cpp
+// Function signature of the completion handler: void(opcua::StatusCode, opcua::Variant&)
+template <typename CompletionToken = opcua::DefaultCompletionToken>
+auto opcua::services::readValueAsync(
```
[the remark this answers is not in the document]
lukasberbuer: Yes, all the *Async functions are coming, hopefully tomorrow :)

Still unsure about the `Node` class. I was thinking about a `NodeAsync` / `AsyncNode` class which only exposes the async functions. Methods could convert between the sync and async node class:

```cpp
Node node = server.getObjectsNode();
// do some sync stuff
NodeAsync nodeAsync = node.toAsync();
// do some async stuff
nodeAsync.toSync().doSyncStuff();
```

I still prefer having a single `Node` class to easily browse all possible functionalities.
Still, the `Server` can't use most of the async functions, so in this regard the async stuff is mainly cluttering.

What do you think?

include/open62541pp/services/detail/ClientService.h
lukasberbuer: I would prefer not to catch the exceptions like Asio is doing it. My biggest concern is, that open62541 is not exception safe and an exception might break a control flow and lead to memory leaks!?

Maybe we just have test this with sanitizers :)

doc/async_model.md
lukasberbuer: You are right, sync calls should not be a problem
But initiating a async call inside a completion handler and waiting for it would block the event loop, correct?

chandryan: Yes, in that case the user would need to take care to do a context switch to another thread first.

include/open62541pp/detail/Result.h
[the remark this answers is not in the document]
chandryan: Hm, yes and no 🤔 

Now I remembered an odd characteristic about OPC UA: According to the [Modeling Best Practices](https://opcfoundation.org/wp-content/uploads/2020/09/OPC-11030-Whitepaper-UA-Modeling-Best-Practices-1.00.00.pdf), application-specific errors are to be transmitted in the following manner:

- Statuscode Uncertain
- The output is only partly valid (at least the first output value is valid, returning an application-specific error code)

In that regard the Result class is a perfect match, as it contains the status code as well as the value. Though I assume this means the following for method calls:

- The Result class might need an additional isUncertain() method for clarity
- The uncertain status code possibility also has to be available in the `useFuture` variant

So in order to support this possibility, I tend towards `std::future<Result<T>>`. What do you think?

doc/async_model.md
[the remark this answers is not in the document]
chandryan: Sounds like a good idea :-) 👍 

I assume one would probably choose to stay with one or the other style, depending on the needs.

include/open62541pp/services/detail/ClientService.h
chandryan: That's a good point and I think it might be safer to set a constraint that the completion handler is `noexcept` then? 
From user perspective I would find that a reasonable constraint, seeing that open62541pp is a wrapper around a C lib :)

doc/async_model.md
lukasberbuer: Just fixed the description in the documentation.

include/open62541pp/services/detail/ClientService.h
lukasberbuer: I totally agree. I see two options:
1. document this
2. enforce noexcept with e.g. a `static_assert(std::is_nothrow_invocable<...>)`

The second option would expect every lambda and function passed as a callback to be explicitly marked as `noexcept`. This is not really practicable so this makes me favor the first option.

include/open62541pp/detail/Result.h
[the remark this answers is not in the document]
lukasberbuer: I didn't know about the best practices paper. Thanks for the link 👍
If we use the `Result` type for service responses, we should use it for the sync functions as well. We could do a non-breaking migration by allowing implicit conversions from `Result<T>` to `T&` but deprecating it. The users code should still work but deprecations will pop up during compiling.

What do you think about `void` results, e.g. from `writeValue`?
1. `void` return type, throw exception on error
2. `Result<void>` return type
3. `StatusCode` return type

lukasberbuer:
Hi @chandryan, implementing all the specialized functions to read/write attributes was quite tedious and error prone. So I decided to make a small code generator to implement them all. They live in a new header `Attribute_highlevel.h` which is include at the bottom of `Attribute.h`.

include/open62541pp/services/detail/ClientService.h
chandryan: Both approaches sound okay to me, I have some questions:

1. Is there a way to safely prevent potential memory leaks in this case, or would you document it sth like "programmer's risk" if they try to catch the exception outside of the callback?
2. Why do you think marking the callbacks explicitly with except is not practicable? For me it would not be practicable if that forbid me to call throwing functions within the callback, but that part still works

include/open62541pp/detail/Result.h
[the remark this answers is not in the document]
chandryan: Good idea with the implicit conversion for compatibility 👍 

Regarding other situations like `writeValue` it looks like also there there might be situations where 2. and 3. make more sense, although the specification is a little fuzzy. The read & write APIs use OPC UA's attribute service set, right? Then this is what I found in the spec:

About Reads:
- [This base page](https://reference.opcfoundation.org/Core/Part4/v104/docs/5.10.2) mentions no uncertain codes, ...
- but the linked [table 178](https://reference.opcfoundation.org/Core/Part4/v104/docs/?r=_Ref127606591) refers to common status codes which do have uncertain codes ...
- and there is a chapter about the returned [DataValue](https://reference.opcfoundation.org/Core/Part4/v104/docs/7.7) which also mentions uncertain status codes

About Writes:
- Similar to reads, [the base page](https://reference.opcfoundation.org/Core/Part4/v104/docs/5.10.4) refers to table 178

I don't understand whether these status codes from the base pages are supposed to represent the only valid subset of table 178. However, probably not, because none of them mention any uncertain codes (even not the [Call service](https://reference.opcfoundation.org/Core/Part4/v104/docs/5.11.2)).

For the sake of supporting uncertain codes in all scenarios and having a consistent Result-oriented API, what do you think about 2. `Result<void>`?
And I think implicit conversions from `Result<void>` to `StatusCode` would also be fine in order to support the 3. proposal based on personal preference?

include/open62541pp/services/detail/ClientService.h
lukasberbuer: 1. An option could be an general-purpose exception handler for both the server and client to allow customization of error handling.
    ```cpp
    using ExceptionHandler = std::function<void(std::exception_ptr)>;
    Server::setExceptionHandler(ExceptionHandler handler);
    ```
    The user decides:
    1. log the exceptions
    2. do nothing
    3. rethrow the exception and catch it around e.g. `Server::run()`. This may cause memory leaks (I have to double check the event loop implementation of open62541)

2. I personally wouldn't like to always specify the noexcept especially for simple lambdas `[](StatusCode code) noexcept {}`... But I could get used to it 😀

chandryan: Your suspicion was right, I just tried it out with a modified client method async example and I did get a bunch of memory leak detections by the sanitizer :-|

I usually try to make it as difficult as possible for a user to stumble into potential pitfalls, that's why I'm a fan of preventing situations ii. and iii., and even no. i. can quickly be missed when an exception is only logged but not actually handled by the user.
Also with a centralized exception handler defined by the library it becomes more difficult to decide (and for the user to understand) when an exception should be caught by the central handler and when it should be forwarded to the user. One could name it something like a "centralized callback exception handler", though. But maybe still difficult to grasp 🤔 

With a callback-specific approach, maybe it helps to provide a default logging wrapper or swallow wrapper for the callable for a user to re-use? But I think that wouldn't help your original intention of not having to specify something extra 😄

lukasberbuer: Thanks for testing! Sadly though...

The exception handler approach disassociates the exception from its source. I guess this would be ok if exceptions are only indicate exceptional conditions, e.g. `std::bad_alloc`. One popular HTTP client library is using a similar approach to approach exceptions in user-defined callbacks: https://github.com/yhirose/cpp-httplib?tab=readme-ov-file#exception-handler

Another idea could be the extension of the `Result` type to allow a result and either a status code or an exception pointer. This is similar to Boost's Outcome design of [outcome::result<T, EC>](https://ned14.github.io/outcome/tutorial/essential/result/) (with error code) and [`outcome::outcome<T, EC, EP>`](https://ned14.github.io/outcome/tutorial/essential/outcome/) (with error code or exception pointer).
I really like binding the exception pointer to the result, but this results in an overhead of +8 bytes.

chandryan: Is there a a distinction between exceptions such as std::bad_alloc and user-defined exceptions? I used to work more without exceptions in the past, so I'm still quite new to that field :-)

I don't have experience working with HTTP, but looking at the protocol I think there are differences in a HTTP scenario compared to OPC UA: HTTP is stateless, so if I understood it right, then from a server perspective an exception handler has the complete context information if it also receives the initial request, as is the case with the handler in cpp-httplib. In comparison OPC UA is stateful, which means even a hypothetical inclusion of the initial request in an exception handler could not serve as full context information.

I guess 8 Bytes, which only add up to the stack in most situations, wouldn't make a perceivable difference. Though I'm not sure if I'm following your idea about the Result type right 🤔 
If you extend the Result type to hold an exception, that exception could only be coming from the TransformResponse handler but not the CompletionHandler, right?
The TransformResponse handler looks like something internal to open62541pp and I don't see how it could throw an exception, because you are only wrapping already-allocated data from the C stack I think? The completion handler running the user code on the other hand would still throw as before.

chandryan: Yes I agree, it's tedious to include exceptions in the Result only for such very rare edge cases :(

If you want to be on the safe side and consider exceptions such as `std::bad_alloc` in the TransformResponse, what about catching the exception and representing it as `UA_STATUSCODE_BADOUTOFMEMORY`? That status code seems to be only used in a server context (as per internet opinion), but open62541 already uses it on the client side in a couple of places.
Although I have no idea how much of an edge case `std::bad_alloc` is anyway - I can hardly imagine a system that runs with dynamic heap allocations and then would not voluntarily or involuntarily crash/restart anyway when it runs out of memory.

lukasberbuer: Yes good idea, just map `std::bad_alloc` to `UA_STATUSCODE_BADOUTOFMEMORY` 👍 

So let's try to enforce the user callbacks to be noexcept. We can later add simple wrapper to make throwing callbacks into noexcept logging callbacks as you suggested, e.g.:

```cpp
template <typename F>
class LogException {
public:
    explicit LogException(F&& func)
        : func_(std::forward<F>(func)) {}

    template <typename... Args>
    auto operator()(Args&&... args) noexcept {
        try {
            return func_(std::forward<Args>(args)...);
        } catch (const std::exception& e) {
            std::cout << e.what() << std::endl;
        }
    }

private:
    F func_;  
};

// noexcept callback
services::readNodeClassAsync(client, id, [](StatusCode code, NodeClass nodeClass) noexcept {
    // ...
});

// wrapped callback
services::readNodeClassAsync(client, id, LogException{[](StatusCode code, NodeClass nodeClass) {
    // ...
}});
```

lukasberbuer:
@chandryan, please check out the idea drafted in #179. This may allow us the ASIO-like approach.

include/open62541pp/services/detail/ClientService.h
lukasberbuer: The `ExceptionCatcher` is now integrated and allows us the Asio-like approach of propagating the exceptions to the `run`/`runIterate` methods: https://github.com/open62541pp/open62541pp/pull/111/commits/6f684c005bfac902dc08fffda25632acf737b27b

I added tests to check the behavior of exceptions thrown in both the transform function and the user-defined callback: https://github.com/open62541pp/open62541pp/blob/6f684c005bfac902dc08fffda25632acf737b27b/tests/ClientService.cpp

doc/async_model.md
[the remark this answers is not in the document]
lukasberbuer: The async `Node` is outside of the scope of this PR. Let us continue the discussion in a dedicated issue / PR.

include/open62541pp/detail/Result.h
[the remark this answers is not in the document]
lukasberbuer: We didn't come to a conclusion if we should expose the `Result` class and use it for async callbacks and returned futures yet.

I think that the error handling could be simplified and improved with the `Result` class. Also the sync functions would benefit from this approach. But I don't see the `Result` class to be production-ready now. We might have to iterate on the interface to make it work smoothly. Therefore I propose to stay with the current design, use, test and improve the `Result` class internally and eventually use it in the future as part of the public API.

We can easily adapt to the `Result` class for asynchronous operations and allow different callback signatures:

```cpp
// now
void(StatusCode code);  // void
void(StatusCode code, T& result);  // non-void

// maybe later
void(Result<void> result);  // void
void(Result<T>& result);  // non-void
```

Do you agree @chandryan?

lukasberbuer:
To finalize this PR, we need async implementations for following service sets:
- [ ] Subscription
- [ ] MonitoredItem
- [ ] View

include/open62541pp/detail/Result.h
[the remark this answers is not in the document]
chandryan: Yes that sounds good, also seeing how big this pull request has become 😀 
Maybe I can take a look at making it more mature in a separate PR. Is there anything specific you have in mind, which is missing in the current implementation?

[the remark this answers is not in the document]
lukasberbuer: Totally, this is getting a monster 😀
I am currently doing a greater refactoring, because we need to move all the contexts of server, client, monitored items, subscriptions, methods to a public header (detail). Only then, we can implement the templated async functions for the subscriptions and monitored items.
Let me finish those refactorings and then discuss how to continue. From there on, we should iterate in smaller steps / PRs.
Do you have something specific in mind, what you would like to work on?

lukasberbuer:
Hi @chandryan, from my perspective this giant PR now could be merged into master to be continued in smaller PRs. Do you agree?
I worked on following topics the last days:
- Move implementation details needed by the async implementation for the Subscription and MonitoredItems services to public (detail) headers
- Add generic `AsyncServiceAdapter` to use any async open62541 with completion tokens. This will be needed for specific async functions like `UA_Client_MonitoredItems_createDataChanges_async`
- Tests for the added async functions. I tried hard to minimize code duplication. The best idea I got, was to use templated tests with `Server`, `Client` and `Async<Client>`. `Async<T>` is just a simple trait the indicates to use async functions. In the tests, a simple check `isAsync<T>` can be used to choose either sync or async functions. This introduces some `if constexpr` checks, but I think it is still readable/understandable.

include/open62541pp/detail/Result.h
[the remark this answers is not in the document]
chandryan: I was thinking about focussing on the Result class, as you said "We might have to iterate on the interface to make it work smoothly". From the first look I couldn't spot anything that comes to my mind which I don't like about the interface, though I could give it some more time to think about it.

PR #118, opened 2023-10-31
fix: return output node id

### LINKED ISSUE #116: Using random NodeIds for Variables and Objects is not possible
First of all: Thanks for this great library!
Using plain open62541 in C++ is quite painful and this project makes it so much easier.
However, I have found a small Problem:

In plain open62541 it is possible to let the library select a node id when adding objects or variables with functions like `UA_Server_addObjectNode` by passing a node id with 0 as numeric identifier (See https://www.open62541.org/doc/master/server.html#node-addition-and-deletion).

However the following Code throws a `BadParentNodeIdInvalid`
``` C++
#include "open62541pp/open62541pp.h"

int main() {
    opcua::Server server;

    // Add a variable node to the Objects node
    opcua::Node parentNode = server.getObjectsNode();
    opcua::Node object = parentNode.addObject({1, 0}, "DummyObject");
    opcua::Node myIntegerNode = object.addVariable({1, 1000}, "TheAnswer");
    // Write some node attributes
    myIntegerNode.writeDisplayName({"en-US", "The Answer"})
        .writeDataType(opcua::DataTypeId::Int32)
        .writeValueScalar(42);

    server.run();
}
```
The reason for this is, that `parentNode.addObject()` creates its return value (`Node`) by using the passed node id, and not the actual node id of the created object. 

It is possible to retrieve the actual node id by using the last parameter (`UA_NodeId *outNewNodeId`) of `UA_Server_addObjectNode`. 

So my suggestion is to use this to return a `Node` that is constructed with the actual node id. 
I created a quick draft of this in https://github.com/open62541pp/open62541pp/compare/master...lbckmnn:open62541pp:feat/random_node_ids and can do a proper PR with tests for this if welcome.

What do you think?

lukasberbuer:
Fixes #116.

### DIFF
--- a/include/open62541pp/Node.h
+++ b/include/open62541pp/Node.h
@@ -70,8 +70,10 @@ public:
         const ObjectAttributes& attributes = {},
         const NodeId& referenceType = ReferenceTypeId::HasComponent
     ) {
-        services::addFolder(connection_, nodeId_, id, browseName, attributes, referenceType);
-        return {connection_, id};
+        NodeId resultingId = services::addFolder(
+            connection_, nodeId_, id, browseName, attributes, referenceType
+        );
+        return {connection_, resultingId};
     }
 
     /// @copydoc services::addObject
@@ -82,10 +84,10 @@ public:
         const NodeId& objectType = ObjectTypeId::BaseObjectType,
         const NodeId& referenceType = ReferenceTypeId::HasComponent
     ) {
-        services::addObject(
+        NodeId resultingId = services::addObject(
             connection_, nodeId_, id, browseName, attributes, objectType, referenceType
         );
-        return {connection_, id};
+        return {connection_, resultingId};
     }
 
     /// @copydoc services::addVariable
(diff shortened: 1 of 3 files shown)


lukasberbuer:
Hi @lbckmnn, I don't have the permission to push to your branch. We only need to fix the inline functions:

```diff
--- a/include/open62541pp/services/NodeManagement.h
+++ b/include/open62541pp/services/NodeManagement.h
@@ -47,7 +47,7 @@ NodeId addObject(
  * @ingroup NodeManagement
  */
 template <typename T>
-inline void addFolder(
+inline NodeId addFolder(
     T& serverOrClient,
     const NodeId& parentId,
     const NodeId& id,
@@ -55,7 +55,7 @@ inline void addFolder(
     const ObjectAttributes& attributes = {},
     const NodeId& referenceType = ReferenceTypeId::HasComponent
 ) {
-    addObject(
+    return addObject(
         serverOrClient,
         parentId,
         id,
@@ -88,14 +88,14 @@ NodeId addVariable(
  * @ingroup NodeManagement
  */
 template <typename T>
-inline void addProperty(
+inline NodeId addProperty(
     T& serverOrClient,
     const NodeId& parentId,
     const NodeId& id,
     std::string_view browseName,
     const VariableAttributes& attributes = {}
 ) {
-    addVariable(
+    return addVariable(
         serverOrClient,
         parentId,
         id,
```

Can you make these changes?

lbckmnn:
Done :+1:

PR #147, opened 2024-01-02
feat: async services (continued)

chandryan:
Continuation of PR https://github.com/open62541pp/open62541pp/pull/111 with the following feature additions on client side:
- Support of pure callback-based async function calls

The callback-based API shares the same *Async suffix as the other functions, but has an additional callback parameter and no return value.
The response callback has a dedicated function parameter *code* for BadStatus occurrences that could occur during in the context of internal response processing, because we don't have a promise to carry it within. A good option would be std::expected, though that is a feature only available from C++23.


chandryan:
Hi @lukasberbuer!

I would like to contribute to open62541pp in the context of supporting asynchronuous method calls :-)
I'm particularly interested in supporting async method calls with a callback-only API (see added code in this PR).

As you have already provided a lot of work, I wanted to ask what you think about the callback-only API and whether I can help filling gaps (e.g. test cases, documentation?).

Some background info about the implementation:
It's a quite basic version of ClientServiceAsync::sendRequest with some small modifications. I would have liked to get a common base to be re-used for ClientServiceAsync::sendRequest, though that turned out to be more complicated than I thought. I couldn't find a way to make sure the promise would be deleted in all error scenarios without making it part of the context, which is already set within ClientServiceAsync::sendRequestCallbackOnly. Also the program flow is slightly different, as without a promise there is no way to easily forward errors to the user-provided callback. 

Kind Regards

lukasberbuer:
Hi @chandryan, great timing and great feature! I would appreciate your support implementing a callback-based async API.

I was reading the great paper "A universal model for asynchronous operations" by Christopher Kohlhoff, the author of ASIO:
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3747.pdf
Another good documentation about this model: https://www.boost.org/doc/libs/1_84_0/doc/html/boost_asio/overview/model/completion_tokens.html

The current (in-work) future-based async API is quite limited and I would like to adapt this universal approach of passing `CompletionTokens` as the last parameter to async functions:
1. Pass callbacks as a completion token
2. Pass special tokens:
   - `opcua::useFuture` to return a future object with the result
   - `opcua::deferred` to return a function object for lazy execution
   - ..., maybe in the future `opcua::useAwaitable` for coroutine support :)  

This might result in quite big changes and refactoring so I would like to move in small steps:
- Implement the callback-based overload for e.g. call (this PR #147)
- Unify `ClientService` to handle both approaches
- Move two overloads to a single templated `callAsync` method
- Adapt this approach for other async functions

Currently I see following challenges:
- `ClientService` is an hidden implementation detail. If we implement the universal async model with the templated `CompletionToken` parameter, `ClientService` must be moved to a public header. `ClientService` on the other hand relies on the `open62541/client.h` header which is also private in open62541pp.. I have to check if the compile times are getting worse if more implementations are exposed in header files. If not, maybe the whole library can be made header-only 🤔

I will add some remarks in the code. Overall the implementation looks solid. Thanks!!

include/open62541pp/Client.h
```diff
@@ -146,7 +146,7 @@ class Client {
      */
     void runIterate(uint16_t timeoutMilliseconds = 1000);
     /// Run the client's main loop by. This method will block until Client::stop is called.
-    void run();
+    void run(uint16_t timeoutMilliseconds = 1000);
```
lukasberbuer, line 149: What is your reason for this change? If we need this, let's make a new PR for this.

tests/Services.cpp
```diff
@@ -20,8 +20,10 @@
 using namespace opcua;
 using namespace std::literals::chrono_literals;
 
-template <typename F>
-constexpr auto syncWrapper(Client& client, F&& asyncClientFunction) {
+template <typename... Args>
+constexpr auto syncWrapper(
+    Client& client, std::future<std::vector<Variant>> (&asyncClientFunction)(Args...)
```
lukasberbuer, line 25: Why do you favor this?

examples/method/client_method_async_callback.cpp
lukasberbuer, line 1: I would prefer to merge this with the `client_method_async` example and just describe both approaches in the comments.

tests/Services.cpp
lukasberbuer: This should also work for other async functions in the `services` namespace. We shouldn't hardcode this to the `callAsync` signature.

examples/method/client_method_async_callback.cpp
```diff
@@ -0,0 +1,43 @@
...
+    auto objNode = client.getObjectsNode();
+    auto greetMethodNode = objNode.browseChild({{1, "Greet"}});
+
+    // Run client event loop in separate thread
+    auto clientThread = std::thread([&] { client.run(0); });
+
+    // Set up a callback with a promise to determine when to exit the example.
+    std::promise<void> promise;
+    auto future = promise.get_future();
+    auto callback = [&promise](UA_StatusCode code, std::vector<opcua::Variant> output) {
```
lukasberbuer, line 22: To distinguish this from the future approach, maybe just `std::cout` the results in the example?

include/open62541pp/Client.h
chandryan: I experienced a delay when stopping the client and tried to adjust the timeout, though I now realized it means runIterate will be called in a polling-like fashion in very short intervals without it. I'll revert this change, as there seems to be no notifying mechanism for stopping the client in open62541.

tests/Services.cpp
chandryan: I had to be more precise for the function signature to support both callAsync signatures. Though although the template parameter of the original syncWrapper looks generic, the implementation does expect the function to return a future.

We could change the signature of the callAsync function with the callback variant to something like `callAsyncWithCallback/callAsyncCallback`. This would make the distinction to the `call/callAsync` pair of the services more clear, then also `syncWrapper` for this variant could be renamed to something like `syncWrapperWithCallback/syncWrapperCallback`.

Do you prefer this solution? If yes, what is your preferred function name?

examples/method/client_method_async_callback.cpp
chandryan: Moved the callback example to `client_method_async` and removed `client_method_async_callback`.

examples/method/client_method_async_callback.cpp
chandryan: You mean std::cout in the callback function?
I have changed both examples to provide identical functionality, so I have changed the promise to also contain the output. Though I'm also fine with changing it back to the promise<void> and a printout in the callback function.

include/open62541pp/Client.h
chandryan: Reverted the changes.

examples/method/client_method_async_callback.cpp
lukasberbuer: Yes, just a callback with `std::cout` outputs results of the method call.
In my opinion, the manual `std::future` approach is very specific and shouldn't used in practice (because we have the other overload returning already a `std::future`).
We should try to make the examples as easy as possible, that's why I would prefer just a simple print in the callback.

tests/Services.cpp
lukasberbuer: `callAsyncWithCallback/callAsyncCallback` sounds good! 👍
This will be a temporary solution. Later, when we have a generic async model, we don't have to test each async model separately. Instead, we should have tests for the generic async handling and one simple test for the actual implementation like `callAsync`.

chandryan: Okay, I have changed it to callAsyncWithCallback for now :-)

chandryan:
This PR is ready for a final review now :-)

How do we like to proceed?

**Merging this PR:**
Unfortunately codecov fails at the moment, though to me it looks like the uncovered branches can only be hit with some in-depth unit tests.
Please check again and tell me whether you prefer to merge it to your working branch or to main. By the way, is there any more work you plan in https://github.com/open62541pp/open62541pp/pull/111 ?

**Support for the universal async model:**
I took a look into the paper and it looks like a really nice, clean approach for the final API :-) Shall we create an issue from [your comment ](https://github.com/open62541pp/open62541pp/pull/147#issuecomment-1875113773) and do the next steps in further pull requests?

lukasberbuer:
Hi @chandryan, looks good, thank you for your work!
I will push some minor changes to make it more generic and looking forward for your feedback. Then let's merge it into the working branch and continue with the other topics.

The code coverage is fine, just ignore the pedantic Codecov actions for now :)

lukasberbuer:
Hi @chandryan, I tried to refactor it to be as generic as possible. Now, the `callImpl` just takes a `CompletionHandler` flag which triggeres the actual execution:
- sync with `UseSync` flag
- async with future with `UseFuture` flag
- async with callback with callable as completion handler

```cpp
template <typename CompletionHandler>
static auto callImpl(
    Client& client,
    const NodeId& objectId,
    const NodeId& methodId,
    Span<const Variant> inputArguments,
    CompletionHandler&& completionHandler
) {
    UA_CallMethodRequest item = createCallMethodRequest(objectId, methodId, inputArguments);
    UA_CallRequest request{};
    request.methodsToCall = &item;
    request.methodsToCallSize = 1;
    return sendRequest<UA_CallRequest, UA_CallResponse>(
        client,
        request,
        [](UA_CallResponse& response) {
            return getOutputArguments(getSingleResultFromResponse(response));
        },
        std::forward<CompletionHandler>(completionHandler)
    );
}

template <>
std::vector<Variant> call(
    Client& client,
    const NodeId& objectId,
    const NodeId& methodId,
    Span<const Variant> inputArguments
) {
    return callImpl(client, objectId, methodId, inputArguments, UseSync{});
}

std::future<std::vector<Variant>> callAsync(
    Client& client,
    const NodeId& objectId,
    const NodeId& methodId,
    Span<const Variant> inputArguments
) {
    return callImpl(client, objectId, methodId, inputArguments, UseFuture{});
}

void callAsync(
    Client& client,
    const NodeId& objectId,
    const NodeId& methodId,
    Span<const Variant> inputArguments,
    AsyncCallback<std::vector<Variant>> callback
) {
    return callImpl(client, objectId, methodId, inputArguments, std::move(callback));
}
```

This results in the most effective code reuse in my opinion. What do you think?
Btw, I started to play around with a very simple implementation of the universal model for async operations: https://godbolt.org/z/roMz17M44

chandryan:
> Hi @chandryan, I tried to refactor it to be as generic as possible. Now, the `callImpl` just takes a `CompletionHandler` flag which triggeres the actual execution:
> 
> * sync with `UseSync` flag
> * async with future with `UseFuture` flag
> * async with callback with callable as completion handler
> 
> ```c++
> template <typename CompletionHandler>
> static auto callImpl(
>     Client& client,
>     const NodeId& objectId,
>     const NodeId& methodId,
>     Span<const Variant> inputArguments,
>     CompletionHandler&& completionHandler
> ) {
>     UA_CallMethodRequest item = createCallMethodRequest(objectId, methodId, inputArguments);
>     UA_CallRequest request{};
>     request.methodsToCall = &item;
>     request.methodsToCallSize = 1;
>     return sendRequest<UA_CallRequest, UA_CallResponse>(
>         client,
>         request,
>         [](UA_CallResponse& response) {
>             return getOutputArguments(getSingleResultFromResponse(response));
>         },
>         std::forward<CompletionHandler>(completionHandler)
>     );
> }
> 
> template <>
> std::vector<Variant> call(
>     Client& client,
>     const NodeId& objectId,
>     const NodeId& methodId,
>     Span<const Variant> inputArguments
> ) {
>     return callImpl(client, objectId, methodId, inputArguments, UseSync{});
> }
> 
> std::future<std::vector<Variant>> callAsync(
>     Client& client,
>     const NodeId& objectId,
>     const NodeId& methodId,
>     Span<const Variant> inputArguments
> ) {
>     return callImpl(client, objectId, methodId, inputArguments, UseFuture{});
> }
> 
> void callAsync(
>     Client& client,
>     const NodeId& objectId,
>     const NodeId& methodId,
>     Span<const Variant> inputArguments,
>     AsyncCallback<std::vector<Variant>> callback
> ) {
>     return callImpl(client, objectId, methodId, inputArguments, std::move(callback));
> }
> ```
> 
> This results in the most effective code reuse in my opinion. What do you think? Btw, I started to play around with a very simple implementation of the universal model for async operations: https://godbolt.org/z/roMz17M44

Had a quick look and it looks very nice, thanks :-) Just a short heads up, I'll be AFK next week and take a deeper look when I'm back.

lukasberbuer:
Let's continue our work in the working branch `async-services` before this PR gets too big.

Issue #149, opened 2024-01-09
Allow passing a logger to the server during construction


lbckmnn:
It is currently not possible to pass a custom logger to the server during construction. 
Because the constructor does some configuration, some log messages get printed to the console.

See the following example:
``` C++
#include <iostream>

#include "open62541pp/open62541pp.h"

int main() {
    opcua::Server server(4840 /* port */);
    server.setLogger([](auto level, auto category, auto msg) {
        std::cout << "[from custom logger][" << opcua::getLogLevelName(level) << "] "
                  << "[" << opcua::getLogCategoryName(category) << "] " << msg << std::endl;
    });
    server.run();
}
```

The console output for this is:
```
[2024-01-09 11:23:09.923 (UTC+0100)] warn/server	AccessControl: Unconfigured AccessControl. Users have all permissions.
[2024-01-09 11:23:09.923 (UTC+0100)] info/server	AccessControl: Anonymous login is enabled
[2024-01-09 11:23:09.923 (UTC+0100)] info/server	AccessControl: x509 certificate user authentication is enabled
[2024-01-09 11:23:09.923 (UTC+0100)] warn/server	Username/Password Authentication configured, but no encrypting SecurityPolicy. This can leak credentials on the network.
[2024-01-09 11:23:09.923 (UTC+0100)] warn/userland	AcceptAll Certificate Verification. Any remote certificate will be accepted.
[from custom logger][info] [network] TCP network layer listening on opc.tcp://[myhost]:4840/
```
Only the last message is from the custom logger and can therefore be controlled. 
The AccessControl seems to be configured in the constructor of the server already.
Would it be possible to pass the logger during construction? 
I don't really care for the first messages but they pollute the console.

lukasberbuer:
Good idea 👍 And thanks for the PR. I will have a look at it.

PR #150, opened 2024-01-09
feat: constructor for Server which takes a logger

lbckmnn:
handles #149

### DIFF
--- a/include/open62541pp/Server.h
+++ b/include/open62541pp/Server.h
@@ -80,6 +80,7 @@ public:
 #endif
 
     /// Set custom logging function.
+    /// Does nothing if the passed function is empty or a nullptr.
     void setLogger(Logger logger);
 
     /// Set custom access control.
--- a/include/open62541pp/Server.h
+++ b/include/open62541pp/Server.h
@@ -45,8 +45,10 @@ public:
      *
      * @param port Port number
      * @param certificate Optional X.509 v3 certificate in `DER` encoded format
+     * @param logger Custom log function. If the passed function is empty, the default logger is
+     * used.
      */
-    explicit Server(uint16_t port = 4840, ByteString certificate = {});
+    explicit Server(uint16_t port = 4840, ByteString certificate = {}, Logger logger = nullptr);
 
 #ifdef UA_ENABLE_ENCRYPTION
     /**
--- a/src/Server.cpp
+++ b/src/Server.cpp
@@ -148,10 +148,20 @@ static void applyDefaults(UA_ServerConfig* config) {
 
 Server::Server(uint16_t port, ByteString certificate, Logger logger)
     : connection_(std::make_shared<Connection>(*this)) {
+    // The logger should be set as soon as possible, ideally even before UA_ServerConfig_setMinimal.
+    // However, the logger gets overwritten by UA_ServerConfig_setMinimal() in older versions of
+    // open62541. The best we can do in this case, is to first call UA_ServerConfig_setMinimal and
+    // then setLogger.
+
+#if UAPP_OPEN62541_VER_GE(1, 1)
     setLogger(std::move(logger));
+#endif
     const auto status = UA_ServerConfig_setMinimal(
         getConfig(this), port, certificate.empty() ? nullptr : certificate.handle()
     );
+#if !(UAPP_OPEN62541_VER_GE(1, 1))
+    setLogger(std::move(logger));
+#endif
     detail::throwOnBadStatus(status);
     applyDefaults(getConfig(this));
     setAccessControl(std::make_unique<AccessControlDefault>());
--- a/src/Server.cpp
+++ b/src/Server.cpp
@@ -146,8 +146,9 @@ static void applyDefaults(UA_ServerConfig* config) {
 #endif
 }
 
-Server::Server(uint16_t port, ByteString certificate)
+Server::Server(uint16_t port, ByteString certificate, Logger logger)
     : connection_(std::make_shared<Connection>(*this)) {
+    setLogger(std::move(logger));
     const auto status = UA_ServerConfig_setMinimal(
         getConfig(this), port, certificate.empty() ? nullptr : certificate.handle()
     );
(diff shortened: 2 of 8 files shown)


include/open62541pp/Server.h
```diff
@@ -147,6 +158,7 @@ class Server {
     ServerContext& getContext() noexcept;
 
 private:
+    void init(uint16_t port, ByteString certificate);
```
lukasberbuer, line 161: I would like to avoid a new (private) member function. A simple free function should work (see below).

src/Server.cpp
```diff
@@ -148,6 +148,16 @@ static void applyDefaults(UA_ServerConfig* config) {
...
+    init(port, std::move(certificate));
+}
+
+Server::Server(Logger logger, uint16_t port, ByteString certificate)
+    : connection_(std::make_shared<Connection>(*this)) {
+    setLogger(std::move(logger));
+    init(port, std::move(certificate));
+}
+
+void Server::init(uint16_t port, ByteString certificate) {
```
lukasberbuer, line 160: This only state the member function accesses in the `ServerConfig*`. Lets make it a static free function and call it in both constructors. E.g.:

```cpp
inline static setMinimal(ServerConfig* config, uint16_t port, const ByteString& certificate) {
   // ...
}
```

And just call it with `setMinimal(getConfig(this), ...)`.

lukasberbuer (changes requested):
Thanks for your contribution - looks good 👍
I just pointed out a small refactoring...

src/Server.cpp
lbckmnn: makes perfect sense  👍

However, should the function really be named `setMinimal` ?

Besides the `UA_ServerConfig_setMinimal()`, both constructor currently also set some default subscription settings (`applyDefaults()`) and configure default access control (`setAccessControl()`).

I would either only put the `UA_ServerConfig_setMinimal` in the static function and leave the other two calls in both constructors, or maybe name the function that does all of that e.g. `initializeServer()`. 
What do you think?

lukasberbuer: Naming is hard 🤔 But you are right, `setMinimal` is a weak choice.
Maybe we go with `initializeServer` or `initServerConfig`. If necessary, we later overload this e.g. for other constructors (`UA_ServerConfig_setDefaultWithSecurityPolicies`).

lbckmnn: done 👍

lukasberbuer: Looks good! 👍

lukasberbuer (approved):

lukasberbuer:
I just had the idea if we can merge both constructors for the sake of simplicity:

```cpp
explicit Server(
    uint16_t port = 4840,
    ByteString certificate = {},
    Logger logger = nullptr  // <-- new parameter
);
```

The only downside I can think of is that you must define the port if using a custom logger `Server(4840, {}, myLogger)`.
What's your take on this @lbckmnn?

lukasberbuer:
I see 🤔 What do we expect, if the user passes a `nullptr` or empty function to the `setLogger` method?
1. disable logging
2. do nothing

Both makes sense to me, it just needs to be documented.

### Option 1

As you said, `std::optional` could be a solution.

### Option 2

We could change the `CustomLogger` implementation to do just nothing if a empty function is provided.

```cpp
void CustomLogger::setLogger(Logger logger) {
    // new
    if (!logger) {
        return;
    }
    // old
    if (nativeLogger_.clear != nullptr) {
        nativeLogger_.clear(nativeLogger_.context);
        nativeLogger_.context = nullptr;
    }
    logger_ = std::move(logger);
    nativeLogger_.log = log;
    nativeLogger_.context = this;
    nativeLogger_.clear = nullptr;
}
```

If logging should be disabled, just do:
```cpp
server.setLogger([](auto&&...) {});
```

I favor option 2. What about you @lbckmnn?

lbckmnn:
I also think Option 2 is in general the better one. 

However, it would break code that depends on the old behavior. 
It is even used in one example:
https://github.com/open62541pp/open62541pp/blob/327f2916a3da795d6d0c1621cbf7b22362efa5f0/examples/client_find_servers.cpp#L38

Do you think that this is acceptable?

lukasberbuer:
I think that's acceptable - let's do it 👍 I will mention it in the CHANGELOG as a potential breaking change.
Will you provide the changes?

lbckmnn:
> I think that's acceptable - let's do it 👍 I will mention it in the CHANGELOG as a potential breaking change. Will you provide the changes?

Perfect 👍
Yes, I will do it in the evening.

lbckmnn:
The ci for v1.0.6 is failing because the logger gets overwritten in `UA_ServerConfig_setMinimal()`
I believe [this](https://github.com/open62541/open62541/commit/b0506d3f7cc93e18bb2c4709ea4dfe51bdb3b8f8) is the commit that fixed that.

So I think we should just check the version and set the logger after UA_ServerConfig_setMinimal on open62541 versions prior to 1.1 and otherwise set the logger first.

src/Server.cpp
```diff
@@ -146,11 +146,22 @@ static void applyDefaults(UA_ServerConfig* config) {
...
+#if UAPP_OPEN62541_VER_GE(1, 1)
+    setLogger(std::move(logger));
+#endif
     const auto status = UA_ServerConfig_setMinimal(
         getConfig(this), port, certificate.empty() ? nullptr : certificate.handle()
     );
+#if !(UAPP_OPEN62541_VER_GE(1, 1))
+    setLogger(std::move(logger));
+#endif
     detail::throwOnBadStatus(status);
```
lukasberbuer, line 165: This should be called directly after `UA_ServerConfig_setMinimal`.

lukasberbuer (approved):

lukasberbuer:
Thank you for your contribution @lbckmnn. Very clean, good comments, tested, ... 👍

src/Server.cpp
lbckmnn: Ah yes, of course :)

PR #155, opened 2024-01-15
feat: add logger to constructor of Client

lbckmnn:
Allows constructing the Client directly with a custom logger. 
See #149  and #150


lukasberbuer:
Thank you @lbckmnn 👍

PR #163, opened 2024-01-16
feat!: Bitmask type to specify bitmasks with enums/ints

lukasberbuer:
- `Bitmask<T>` to allow both enum classes and native enums/ints to define bitmasks:
   
   ```cpp
   // construct with scoped enums
   Bitmask<NodeClass> mask = NodeClass::Variable | NodeClass::Object;
   // construct with unscoped (native) enums or ints
   Bitmask<NodeClass> mask = UA_NODECLASS_VARIABLE | UA_NODECLASS_OBJECT;
   ```

- `IsBitmaskEnum<T>` trait to enable bitwise operators of enum classes:

   ```cpp
   // define enum (class)
   enum class Access {
       Read  = 1 << 0,
       Write = 1 << 1,
   };
   
   // allow bitwise operations
   template <>
   struct IsBitmaskEnum<Access> : std::true_type {};
   
   // use bitwise operations
   Access mask = Access::Read | Access::Write;
   ```

### CHANGES

- Add `AccessLevel`, `WriteMask` and `EventNotifier` enum classes
- Use `Bitmask<AccessLevel>` instead of `uint8_t` with implicit conversions to/from `uint8_t`
- Use `Bitmask<WriteMask>` instead of `uint32_t` with implicit conversions to/from `uint32_t`
- Implicit conversion from `Bitmask<T>` to underlying integer are deprecated and will be made explicit in the future.
  Please migrate to `Bitmask<T>::get()`.

### BREAKING CHANGES

- Virtual function `AccessControl::getUserRightsMask` returns `Bitmask<WriteMask>` instead of `uint32_t`
- Virtual function `AccessControl::getUserAccessLevel` returns `Bitmask<AccessLevel>` instead of `uint8_t`


lukasberbuer:
@johanneskopf, you have been using the `AccessControl` plugin which is affected by this PR. Do you have any remarks?

Issue #164, opened 2024-01-17
Setting an array with type bool does not work (assertion error or segfault)


lbckmnn:
# Problem
The very unfortunate specialization for `std::vector<bool>` is causing some troubles with Variant.setArray():
Consider the following example:

``` C++
int main() {
  Variant variant;
  std::vector<bool> vector{false, true, true, false};
  //variant.setArrayCopy(vector); // does not compile
  variant.setArrayCopy(vector.begin(), vector.end()); // fails with an assertion / segfaults
}
```
This code fails with the assertion:
```
constexpr T opcua::detail::copy(const T&, const UA_DataType&) [with T = std::_Bit_reference; UA_DataType = UA_DataType]:
```
In release build it even fails with a segfault.

# Investigation
The underlying problem is, that accessing an iterator of a `std::vector<bool>` yields a `Bit_reference` and not a bool.
With e.g. an `std::vector<int>` accessing an iterator yields an int.

https://github.com/open62541pp/open62541pp/blob/7e75bd39080832837cce87b09bc076500a783adb/include/open62541pp/types/Variant.h#L428-L436

The lambda in `std::transform` receives a Bit_reference and passes that into `detail::copy`, which expects a bool.

# Other observations
Using an `std::array<bool>` works just fine and `Variant.getArrayCopy` is also unproblematic. 
However, `Variant.setArrayCopy(std::vector<bool>)` does not compile because it is (fortunately) not convertiable to a `Span`

lbckmnn:
I can also take a further look at this and help or provide a fix if wanted.

lukasberbuer:
Great catch and investigation, thank you! 👍

As you pointed out, the problem is the `std::transform` of a `vector<bool>` in the `getArrayCopyImpl` method.
The type of the `value` parameter of the lambda is actually a `std::vector<bool>::reference`: https://en.cppreference.com/w/cpp/container/vector_bool/reference

Luckily, the reference type can be converted to a bool with a simple `static_cast`. So this simple modification should fix the problem:

```cpp
    std::transform(first, last, native.get(), [&](auto&& value) {
        return detail::copy(static_cast<ValueType>(value), dataType);
        //                  ^^^^^^^^^^^^^^^^^^^^^^
    });
```

It shouldn't have any consequences for other types. Did I miss something?
Would you provide a patch with a comment why we changed this?

lbckmnn:
> It shouldn't have any consequences for other types. Did I miss something?

I can't think of any. 

[...]

PR #167, opened 2024-01-22
fix: Variant::setArrayCopy

lbckmnn:
This PR fixes some problems that can be observed, when using one of the following overloads with `std::vector<bool>` due to its specialization.

1. `void setArrayCopy(ArrayLike&& array)`
2. `void setArrayCopy(ArrayLike&& array, const UA_DataType& dataType)`
3. `void setArrayCopy(InputIt first, InputIt last)`
4. `void setArrayCopy(InputIt first, InputIt last, const UA_DataType& dataType)`

See #164 for details.

### DIFF
--- a/include/open62541pp/types/Variant.h
+++ b/include/open62541pp/types/Variant.h
@@ -311,8 +311,14 @@ public:
 
     /// @overload
     template <typename ArrayLike, typename = EnableIfNoSpan<ArrayLike>>
-    void setArrayCopy(ArrayLike&& array) noexcept {
-        setArrayCopy(Span{std::forward<ArrayLike>(array)});
+    void setArrayCopy(ArrayLike&& array) {
+        using ValueType = typename std::remove_reference_t<ArrayLike>::value_type;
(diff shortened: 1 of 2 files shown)


include/open62541pp/types/Variant.h
```diff
@@ -483,7 +495,7 @@ void Variant::setArrayCopyImpl(InputIt first, InputIt last, const UA_DataType& d
     using ValueType = typename std::iterator_traits<InputIt>::value_type;
     const size_t size = std::distance(first, last);
     auto native = detail::allocateArrayUniquePtr<ValueType>(size, dataType);
-    std::transform(first, last, native.get(), [&](auto&& value) {
+    std::transform(first, last, native.get(), [&](const ValueType& value) {
```
lbckmnn, line 498: Instead of casting to ValueType, I just changed the type directly. 

If this is changed back to `auto&&`, the added test will fail.

lukasberbuer: Best solution! 👍

PR #178, opened 2024-01-29
feat: variant policy for Variant::fromScalar, Variant::fromArray

lukasberbuer:
A possible solution for #174.

### DIFF
--- a/include/open62541pp/types/DataValue.h
+++ b/include/open62541pp/types/DataValue.h
@@ -52,16 +52,16 @@ public:
 
     /// Create DataValue from scalar value.
     /// @see Variant::fromScalar
-    template <typename... Args>
+    template <VariantPolicy Policy = VariantPolicy::ReferenceIfPossible, typename... Args>
     [[nodiscard]] static DataValue fromScalar(Args&&... args) {
-        return DataValue(Variant::fromScalar(std::forward<Args>(args)...));
+        return DataValue(Variant::fromScalar<Policy>(std::forward<Args>(args)...));
     }
 
     /// Create DataValue from array.
     /// @see Variant::fromArray
-    template <typename... Args>
+    template <VariantPolicy Policy = VariantPolicy::ReferenceIfPossible, typename... Args>
     [[nodiscard]] static DataValue fromArray(Args&&... args) {
-        return DataValue(Variant::fromArray(std::forward<Args>(args)...));
+        return DataValue(Variant::fromArray<Policy>(std::forward<Args>(args)...));
     }
 
     bool hasValue() const noexcept {
--- a/include/open62541pp/types/Variant.h
+++ b/include/open62541pp/types/Variant.h
@@ -24,15 +24,9 @@ namespace opcua {
 // forward declarations
 class NodeId;
 
-enum class VariantPolicy {
-    CopyOnly,
-    AssignOnly,
-    AssignIfPossible,
-};
-
 namespace detail {
 
-template <VariantPolicy Policy>
+template <typename Policy>
 struct VariantHandler;
 
 }  // namespace detail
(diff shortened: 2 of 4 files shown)


lbckmnn:
This approach has one big problem that becomes apparent, when you try to use the `Variant::fromScalar/Array` functions with a different policy than the default one:
template argument deduction and default arguments doesn't really mix well:

The signature of e.g. the fromArray function is:

```C++
template <typename ArrayLike, VariantPolicy Policy = VariantPolicy::AssignIfPossible>
static Variant fromArray(ArrayLike&& array);
```
To specify another policy, you also have to specify `ArrayLike` instead of letting the compiler deduce it.
And because the value type would normally also be deduces (forwarding reference) you even need to specify the exact value type:

```C++
int main() {
    std::vector<int> vec{1, 2, 3};
    auto variant = opcua::Variant::fromArray<const std::vector<int>&, opcua::VariantPolicy::CopyOnly>(vec);
}
```

I will try to come up with something that allows to just do:
```C++
std::vector<int> vec{1, 2, 3};
auto variant = opcua::Variant::fromArray<opcua::VariantPolicy::CopyOnly>(vec);
```

What do you think of just using dedicated functions?

e.g. fromScalar, fromScalarAssign, fromScalarCopy

lbckmnn:
> I will try to come up with something that allows to just do:

It is possible to swap the arguments and add another overload for defining the default policy like this:

```C++

 /// Create Variant from scalar value using the default policy AssignIfPossible
    template <typename T>
    [[nodiscard]] static Variant fromScalar(T&& value) {
       return fromScalar<VariantPolicy::AssignIfPossible, decltype(value)>(std::forward<T>(value));
    }

    /// Create Variant from scalar value.
    template <VariantPolicy Policy, typename T>
    [[nodiscard]] static Variant fromScalar(T&& value) {
        Variant var;
        detail::VariantHandler<Policy>::setScalar(var, std::forward<T>(value));
        return var;
    }
```

It is then possible to use type deduction:

``` C++
    bool value = true;
    auto variant = Variant::fromScalar<VariantPolicy::CopyOnly>(value);
```

However, this complicates the implementation again and needs additional 6 overloads...

lbckmnn (commented):
see comments

lukasberbuer:
I see. What about providing the policy as a function parameter (tag)? This pattern is also used in the standard for the execution policy of algorithms: https://en.cppreference.com/w/cpp/algorithm/execution_policy_tag_t

```cpp
struct VariantPolicy {
    struct CopyOnly {};
    struct AssignOnly {};
    struct AssignIfPossible {};

    using Default = AssignIfPossiblePolicy;
};

class Variant {
public:
    // ...
    template <typename ArrayLike, typename Policy = VariantPolicy::Default>
    [[nodiscard]] static Variant fromArray(ArrayLike&& array, Policy = VariantPolicy::Default());
    // ...
};

int main() {
    std::vector<int> vec{1, 2, 3};
    auto variant = opcua::Variant::fromArray<const std::vector<int>&>(vec, opcua::VariantPolicy::CopyOnly{});
}
```

Still unsure if those tag types should be available either:
- globally: `opcua::CopyOnly`
- in namespace or struct: `opcua::VariantPolicy::CopyOnly`
- in `Variant` class: `opcua::Variant::CopyOnly`

Additional global variant policy objects could be easily added for easier handling:
```cpp
static constexpr opcua::VariantPolicy::CopyOnly copyOnly{};
```

lbckmnn:
That looks promising. 
I think we can even implement this without the VariantHelper.

> Still unsure if those tag types should be available either:
>
>    - globally: opcua::CopyOnly
>    - in namespace or struct: opcua::VariantPolicy::CopyOnly
>    - in Variant class: opcua::Variant::CopyOnly


Can you think of any other Type where this policy might be useful?

lukasberbuer:
The tag approach resulted in way more code than expected: https://github.com/open62541pp/open62541pp/pull/178/commits/2d55e7e66d17ec44fd546e64189fcabf5bd29986

I still like the very simple approach with the `VariantPolicy` enum. Maybe we should come back to your idea of swapping the template parameters:

```cpp
template <VariantPolicy Policy = VariantPolicy::AssignIfPossible, typename ArrayLike>
static Variant fromArray(ArrayLike&& array);
```

The change would break following code:

```cpp
Variant::fromScalar<int16_t>(11);

// has to be rewritten as
Variant::fromScalar(int16_t{11});
Variant::fromScalar(static_cast<int16_t>(11));
```

I found such snippets in the tests. Although I expect the practical usage to be quite rare. What to you think @lbckmnn?

lbckmnn:
> I found such snippets in the tests. Although I expect the practical usage to be quite rare. What to you think @lbckmnn?

I think that is just fine at the current stage of this project 👍

PR #179, opened 2024-01-30
feat: propagate callback exceptions to event loop run method

lukasberbuer:
A possible solution to handle exception is user-defined callbacks, e.g. for asynchronous operations as discussed in #111.

Idea:
- `Server` and `Client` store an `exception_ptr` pointer internally
- Callbacks are executed inside a catching wrapper, which sets the interal `exception_ptr`
- The `exception_ptr` is checked every iteration of `run` / `runIterate` and throws if an exception was generated

This solves possible memory leaks by throwing exceptions in an exception-unaware context (open62541 / C).


chandryan:
Just storing the exception instead of handling it -> great idea, I like it! 🥳

src/ExceptionHandler.h
```diff
@@ -0,0 +1,71 @@
+#pragma once
+
+#include <exception>
+#include <functional>
+#include <utility>  // exchange
+
+namespace opcua {
+
+class ExceptionHandler {
```
chandryan, line 9: I guess this class would be something like an `ExceptionCatcher` in this context?

lukasberbuer: Or maybe `ExceptionPropagation`?

lukasberbuer: Renamed to `ExceptionCatcher`, good idea! 👍

Issue #180, opened 2024-02-02
Calling methods with no outputs crashes the program


lbckmnn:
Minimal example:

``` C++
#include "open62541pp/open62541pp.h"
#include <iostream>
#include <thread>

int main() {
    opcua::Server server;

    auto objectNode = server.getObjectsNode();
    objectNode.addMethod(
        {1, 1000},
        "empty method",
        [](opcua::Span<const opcua::Variant>, opcua::Span<opcua::Variant>) {
            std::cout << "hello world" << std::endl;
        },
        {},
        {}
    );
    std::thread serverRunner(&opcua::Server::run, &server);
    std::this_thread::sleep_for(std::chrono::milliseconds(100));
    opcua::Client client;
    client.connect("opc.tcp://127.0.0.1:4840");
    client.getObjectsNode().callMethod({1, 1000}, {});
    client.disconnect();
    server.stop();
    serverRunner.join();
}
```

This is the method that eventually calls `UA_client_call()`

https://github.com/open62541pp/open62541pp/blob/1f8222804ce7f7918303ca03e3dba06072aaecc9/src/services/Method.cpp#L45-L72

The problem is the deallocation of the array (`UA_free(array)`) and the way open62541 handles arrays. 

Empty arrays have the size zero and the data pointer `UA_EMPTY_ARRAY_SENTINEL` (which is 0x1).
When the array is empty, `UA_free(0x1)` is called. 

Arrays should be deallocated using `UA_Array_delete(output, outputSize, &UA_TYPES[UA_TYPES_VARIANT]);`

PR #181, opened 2024-02-02
fix: use UA_Array_delete instead of UA_free

lbckmnn:
Arrays should be deallocated using UA_Array_delete.
Otherwise non trivial members are not deleted and empty arrays cause problems. 
See #180


lukasberbuer:
Perfect, thanks 🙏

PR #184, opened 2024-02-05
perf: move rvalue scalars to variant

lukasberbuer:
Avoid copies of rvalue scalars using `Variant::setScalarCopy` (if possible).

As discussed in #174 with @lbckmnn.

### DIFF
--- a/include/open62541pp/types/Variant.h
+++ b/include/open62541pp/types/Variant.h
@@ -297,7 +297,7 @@ public:
      *              The container must implement `begin()` and `end()`.
      */
     template <typename ArrayLike>
-    void setArrayCopy(const ArrayLike& array) {
+    void setArrayCopy(ArrayLike&& array) {
         setArrayCopy(array.begin(), array.end());
     }
 
@@ -307,7 +307,7 @@ public:
      * @param dataType Custom data type.
      */
     template <typename ArrayLike>
-    void setArrayCopy(const ArrayLike& array, const UA_DataType& dataType) {
+    void setArrayCopy(ArrayLike&& array, const UA_DataType& dataType) {
         setArrayCopy(array.begin(), array.end(), dataType);
     }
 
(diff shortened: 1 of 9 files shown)


include/open62541pp/types/Variant.h
```diff
@@ -457,9 +460,15 @@ void Variant::setArrayImpl(
...
-    auto native = detail::allocateUniquePtr<T>(dataType);
-    *native = detail::copy(value, dataType);
+void Variant::setScalarCopyImpl(T&& value, const UA_DataType& dataType) {
+    using ValueType = std::remove_cv_t<std::remove_reference_t<T>>;
+    auto native = detail::allocateUniquePtr<ValueType>(dataType);
+    // https://stackoverflow.com/a/50086839/9967707
+    if constexpr (std::is_lvalue_reference_v<T>) {
+        *native = detail::copy(value, dataType);
+    } else {
+        *native = std::exchange(value, {});
```
lbckmnn, line 470: can't we just move assign here?

include/open62541pp/types/Variant.h
```diff
@@ -507,11 +516,21 @@ struct VariantHandler<VariantPolicy::Copy> {
         var.setScalarCopy(value);
     }
 
+    template <typename T, typename = std::enable_if_t<!std::is_reference_v<T>>>
```
lbckmnn, line 519: yeah, it is really hard to see why this `std::enable_if` is needed, just by looking at the code.

lbckmnn (commented):
Nice idea, but I am not sure if it is worth the extra complexity.

include/open62541pp/types/Variant.h
lukasberbuer: Integral values and simple C `struct`s will just copy if using `std::move`. Zero-initializing the source object prevents maybe a necessary deallocation of the source object.

include/open62541pp/types/Variant.h
lukasberbuer: Totally agree.. let leave it as a draft PR.

PR #208, opened 2024-02-29
refactor: connection registry

lukasberbuer:


lukasberbuer:
Too much overhead with shared/weak pointers and locking...

PR #264, opened 2024-03-27
feat: `Result` monadic operations transform, andThen, orElse

lukasberbuer:
Add monadic operations to `Result<T>` class:
- `Result<T>::transform`
- `Result<T>::andThen`
- `Result<T>::orElse`

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -7,8 +7,9 @@ AccessModifierOffset: -4
 AlignAfterOpenBracket: BlockIndent # version >= 14 required
 # AlignConsecutiveAssignments: true
 # AlignConsecutiveDeclarations: true
-AlignTrailingComments: false
+AlignOperands: DontAlign
 AllowShortFunctionsOnASingleLine: Empty
+AlignTrailingComments: false
 AlwaysBreakTemplateDeclarations: Yes
 BinPackArguments: false
 BinPackParameters: false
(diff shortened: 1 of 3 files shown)


lukasberbuer:
Any remarks @chandryan?

chandryan:
> Any remarks @chandryan?

Looks very nice 🥳

lukasberbuer:
> > Any remarks @chandryan?
> 
> Looks very nice 🥳

Thanks 😊

Now that we have to include `<functional>` and `<type_traits>` in `Result.h`, I don't see any need to keep the separate header `result_util.h`. Do you have any reasons not to merge them?

chandryan:
> > > Any remarks @chandryan?
> > 
> > 
> > Looks very nice 🥳
> 
> Thanks 😊
> 
> Now that we have to include `<functional>` and `<type_traits>` in `Result.h`, I don't see any need to keep the separate header `result_util.h`. Do you have any reasons not to merge them?

I think it depends on whether you want to make `tryInvoke` a public API of open62541pp or leave it in the `detail` namespace, would it provide a benefit for users?
It could be confusing in case open62541pp offers a direct way to register a C callback towards open62541 that returns a uint32 UA_STATUSCODE; then we would have the ambiguity of its return value being interpreted as `Result<uint32_t>`.

If there is no use for it, I'd recommend to keep it in the detail namespace and thus in a separate header file.

PR #266, opened 2024-03-27
build: check std::filesystem support and conditionally enable fs-related features

### LINKED ISSUE #265: Compiler without proper std::fs support
### Discussed in https://github.com/open62541pp/open62541pp/discussions/263

<div type='discussions-op-text'>

<sup>Originally posted by **ErikTempelaarVO** March 27, 2024</sup>
I have a question regarding the compile-time check to determine if std::fs is experimental / supported.

https://github.com/ErikTempelaarVO/open62541pp/blob/7e75bd39080832837cce87b09bc076500a783adb/include/open62541pp/types/Builtin.h#L12

I believe the logic is a bit flawed, because if the compiler doesn't have any support, it assumes full support.
Which is wrong for my compiler-variant, as already mentioned here:
https://github.com/open62541pp/open62541pp/pull/110#issuecomment-1807794052

````
#if !__has_include(<filesystem>) && __has_include(<experimental/filesystem>)
#include <experimental/filesystem>
namespace fs = std::experimental::filesystem;
#else
#include <filesystem>
namespace fs = std::filesystem;
#endif
````

I create a workaround for my compiler here:
https://github.com/open62541pp/open62541pp/commit/e5e8a6c4fefe1ec21751b6181ad35068d7af6cc3

Is it feasible to add an option to the library that disables any use of std::fs?</div>

lukasberbuer:
Support compilers with partial C++17 support, e.g. GCC 7. Closes #265.

Is this a proper solution for you @ErikTempelaarVO?

### DIFF
--- a/doc/Doxyfile.in
+++ b/doc/Doxyfile.in
@@ -75,6 +75,7 @@ PREDEFINED             = UA_ENABLE_DA \
                          UA_ENABLE_QUERY \
                          UA_ENABLE_SUBSCRIPTIONS \
                          UA_ENABLE_SUBSCRIPTIONS_EVENTS \
+                         UAPP_HAS_FILESYSTEM \
                          UAPP_CREATE_CERTIFICATE
 
 # Dot diagrams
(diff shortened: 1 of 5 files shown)


ErikTempelaarVO:
I think this will break for people relying in experimental support?

lukasberbuer:
You are right @ErikTempelaarVO. Just took your logic and only renamed the flag.

PR #384, opened 2024-10-19
feat!: unified API for variants (constructing and setting) and removal of VariantPolicy::ReferenceIfPossible

chandryan:
# Goal

Add a unified API for constructing and setting variants. This merge request includes:

- universal constructors with default copy behavior and a reference tag
- setValue / setValueCopy APIs

Also VariantPolicy::ReferenceIfPossible is removed, as it creates ambiguity about the variant's effective handling of data.


include/open62541pp/types.hpp
```diff
@@ -888,59 +903,87 @@ class Variant : public TypeWrapper<UA_Variant, UA_TYPES_VARIANT> {
...
+        setValue<ConstructorToken::policy>(first, last);
+    }
+
+    template <typename ConstructorToken = ConstructorToken<VariantPolicy::Copy>, typename InputIt>
+    Variant(InputIt first, InputIt last, const UA_DataType& dataType, ConstructorToken /*unused*/) {
+        setValue<ConstructorToken::policy>(first, last, dataType);
+    }
+
+    // TODO:
+    // Do not add fromValue methods? As we the constructors above.
```
chandryan, line 929: Which variant do you prefer @lukasberbuer ?

- Only constructors,
- Only `fromValue` or
- Both constructors and `fromValue?`

include/open62541pp/node.hpp
chandryan, line 1: Do you think it's worth adding `readValue`/`writeValue` methods for the Node class?

I'm not sure as they only expose a subset of functionality and only sync methods. For users going via the `Variant` class seems only a small indirection, while duplicating parts of the `Variant` API into the `Node` class might be more difficult to maintain.

include/open62541pp/types.hpp
```diff
@@ -1090,6 +1184,32 @@ class Variant : public TypeWrapper<UA_Variant, UA_TYPES_VARIANT> {
         return getArrayCopyImpl<T>();
     }
 
+    // Unified API for setting a value.
+    // A template parameter is used to select the variant
+    // policy instead of using additional functions with "Copy" suffix.
+    // Default policy is Copy for more robust behavior (avoid potential dangling references and
```
chandryan, line 1190: What's your thought on using "Copy" as default policy when setting values?
On one hand it is safer due to accidental dangling references; on the other hand I imagine the `Variant` class would be used as a short-living wrapper to encapsulate some data for sending it away in most cases.

chandryan (commented):
@lukasberbuer I think the pull request is now in a mature-enough state to discuss API decisions. Please provide feedback, then I'd take a look into polishing tests and documentation 😄

include/open62541pp/types.hpp
lukasberbuer: The `Variant` interface is already quite complex. I would prefer constructors only.

include/open62541pp/types.hpp
```diff
@@ -875,6 +875,21 @@ enum class VariantPolicy {
     // clang-format on
 };
 
+/**
+ * Token used for selecting the policy when constructing variants.
+ *
+ * TODO:
```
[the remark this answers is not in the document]
lukasberbuer: The STL uses tag types always as the first constructor argument, for example `std::pair` with `std::piecewise_construct_t` or `std::optional` with `std::in_place_t`. We should do that as well.
The default should always be the safe choice which is copying. Therefore we need tag types for referencing / borrowing the provided value.

Naming is hard, but this would be my favorites:
```cpp
struct BorrowTag {};
inline constexpr BorrowTag borrow{};
```

Other naming ideas are:
- `NoCopyTag`
- `NoDeleteTag`

What do you think?

include/open62541pp/node.hpp
lukasberbuer: I don't think so. It's very simple and also expressive enough to just write:
```cpp
opcua::Node node(...);
node.writeValue(opcua::Variant(5));
```

I almost regret adding the shortcuts like `Node::writeValueScalar`, .... It makes the code a bit shorter, but bloats the API of the `Node` class (which is already quite big).

include/open62541pp/types.hpp
lukasberbuer: Copy is the safe choice, therefore it should be the default in my opinion.
I agree, that the `Variant` is usually a short-living object, therefore it makes sense to just borrow the data. But this is an optimization that should be introduced explicitly by the library users.

After some experience with the current policy model, I would like to simplify this. Either the `Variant` owns the data or it doesn't. The `ReferenceIfPossible` policy hides if the provided value is actually borrowed or copied. If a value can not be borrowed, a compile error is the safest choice and makes clear, that an explicit copy is needed.

Do you agree?

lukasberbuer:
Great work @chandryan, thank you 👍

Currently, I see the biggest challenge in keeping the `Variant` interface small and concise. My biggest pains at the moment are:
- The `VariantPolicy` design, especially the `ReferenceIfPossible` policy with its opaque behavior.
- The huge amount of setter methods. Actually, they are not needed anymore, because the `Variant` can be constructed in every possible way as the setter methods allow it. Especially your added constructors make it very easy to use. Performance wise, there is not advantage in using the setters instead of reconstructing the variant.
- Are the ref qualifiers in the getters really safe? If the data is borrowed, the move with rvalue ref qualifiers will move from the origin, right? Is that what you expect?

Changing this will introduce breaking changes, but it might be worth it.

This is how the `Variant` interface could look like with you universal constructors and with a simple policy design with a single tag `BorrowTag` / `borrow`:

```cpp
struct BorrowTag {};
inline constexpr BorrowTag borrow{};

template <typename T>
Variant(T&& value);

template <typename T>
Variant(BorrowTag, T& value) noexcept;

template <typename T>
Variant(T&& value, const UA_DataType& dataType);

template <typename T>
Variant(BorrowTag, T& value, const UA_DataType& dataType) noexcept;

template <typename InputIt>
Variant(InputIt first, InputIt last);

template <typename InputIt>
Variant(InputIt first, InputIt last, const UA_DataType& dataType);

template <typename T>
[[nodiscard]] static Variant fromScalar(T&& value);

template <typename T>
[[nodiscard]] static Variant fromScalar(BorrowTag, T& value) noexcept;

template <typename T>
[[nodiscard]] static Variant fromScalar(T&& value, const UA_DataType& dataType);

template <typename T>
[[nodiscard]] static Variant fromScalar(BorrowTag, T& value, const UA_DataType& dataType) noexcept;

template <typename ArrayLike>
[[nodiscard]] static Variant fromArray(ArrayLike&& array);

template <typename ArrayLike>
[[nodiscard]] static Variant fromArray(BorrowTag, ArrayLike& array) noexcept;

template <typename ArrayLike>
[[nodiscard]] static Variant fromArray(ArrayLike&& array, const UA_DataType& dataType);

template <typename ArrayLike>
[[nodiscard]] static Variant fromArray(BorrowTag, ArrayLike& array, const UA_DataType& dataType) noexcept;

template <typename InputIt>
[[nodiscard]] static Variant fromArray(InputIt first, InputIt last);

template <typename InputIt>
[[nodiscard]] static Variant fromArray(InputIt first, InputIt last, const UA_DataType& dataType);

bool isEmpty() const noexcept;

bool isScalar() const noexcept;

bool isArray() const noexcept;

bool isType(const UA_DataType* dataType) const noexcept;

bool isType(const UA_DataType& dataType) const noexcept;

bool isType(const NodeId& id) const noexcept;

template <typename T>
bool isType() const noexcept;

const UA_DataType* getDataType() const noexcept;

void* data() noexcept;

const void* data() const noexcept;

// really needed?
template <typename T>
auto getValue() noexcept;

// really needed?
template <typename T>
T getValueCopy() const &&;

template <typename T>
T& getScalar() &;

template <typename T>
const T& getScalar() const&;

template <typename T>
T&& getScalar() &&;

template <typename T>
const T&& getScalar() const&&;

template <typename T>
T getScalarCopy() const;

size_t getArrayLength() const noexcept;

Span<const uint32_t> getArrayDimensions() const noexcept;

template <typename T>
Span<T> getArray();

template <typename T>
Span<const T> getArray() const;

template <typename T>
std::vector<T> getArrayCopy() const;
```

The `auto` return type of the `getValue` member function makes it hard to understand. I wonder if `getScalar*` and `getArray*` are enough? What benefits do you see?

include/open62541pp/types.hpp
chandryan: Removed the `fromValue` functions

include/open62541pp/types.hpp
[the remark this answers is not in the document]
chandryan: Indeed naming is hard :D
I have no better idea than what you have already chosen for the variant policy. I don't understand the other names yet (are they all alternatives to a potential `ReferenceTag`?):

- `BorrowTag` makes me think of an actual borrow checker, which we don't have in C++ (yet ^^)
- `NoCopyTag` could mean referencing the data or moving it into the Variant
- `NoDeleteTag` excludes ownership and suggests no copy but seems very indirect

chandryan:
> 
> Changing this will introduce breaking changes, but it might be worth it.
> 
> This is how the `Variant` interface could look like with you universal constructors and with a simple policy design with a single tag `BorrowTag` / `borrow`:
> ...

True, there is quite some potential if breaking changes are acceptable :-)

In that case would `fromArray` and `fromScalar` still be required? I think the constructors should be able to cover all use cases.

chandryan:
> The `auto` return type of the `getValue` member function makes it hard to understand. I wonder if `getScalar*` and `getArray*` are enough? What benefits do you see?

I prefer an API with which I can specify the expected data type via a single parameter, it seems more intuitive and less verbose.

With an API that lets me define all details in the template parameter I can write code like

- `auto value = variant.getValue<int32_t>;` and `auto value = variant.getValue<std::vector<int32_t>>;` instead of
- `auto value = variant.getScalar<int32_t>;` and `auto value = variant.getArray<int32_t>;`.

With the first version I can clearly decide on and see the data type I receive. This is also more universal in case one would like to support any one-dimensional STL containers in the future. You could then for example state `auto value = variant.getValue<std::list<int32_t>>;`. Or in case supporting multi-dimensional arrays becomes a thing you could write `auto value = variant.getValue<std::vector<std::vector<int32_t>>>;`


I agree with `auto` resp. `decltype(auto)` not looking very nice, though. I needed `decltype(auto)` to get the correct return types from all called functions.

chandryan:
> * The huge amount of setter methods. Actually, they are not needed anymore, because the `Variant` can be constructed in every possible way as the setter methods allow it. Especially your added constructors make it very easy to use. Performance wise, there is not advantage in using the setters instead of reconstructing the variant.

I'm cautious with removing setter methods and forcing the user to always re-construct. It seems more fitting for scenarios such as lock guards and I would not rule out that they are needed with some code patterns. For example:

```
Variant var;

if (conditionA)
  var.setValue(4);
else if (conditionB)
  var.setValue(4.0f);
else
  var.setValue("Hello");

node.writeValue(var);
```

lukasberbuer:
> In that case would `fromArray` and `fromScalar` still be required? I think the constructors should be able to cover all use cases.

True. We could keep them for backward compatibility with perfect forwarding to the constructor and add a deprecation.

include/open62541pp/types.hpp
chandryan: Yes I agree, it would be better to drop `ReferenceIfPossible` 👍

lukasberbuer:
> I'm cautious with removing setter methods and forcing the user to always re-construct. It seems more fitting for scenarios such as lock guards and I would not rule out that they are needed with some code patterns. For example:
> 
> ```
> Variant var;
> 
> if (conditionA)
>   var.setValue(4);
> else if (conditionB)
>   var.setValue(4.0f);
> else
>   var.setValue("Hello");
> 
> node.writeValue(var);
> ```

On the other hand re-contructing is the common way in the STL, e.g. with `std::variant`, `std::optional`...
I would argue, that the shown example is not best practice anyway. I could be easily refactored with a factory function or even a immediately invoked lambda. As a result, the created `Variant` can be marked `const` or passed directly to another function as a rvalue reference:

```cpp
const auto createVariant = [&] {
  if (conditionA)
    return Variant(4);
  if (conditionB)
    return Variant(4.0f);
  return Variant("Hello");
};

const auto var = createVariant();
node.writeValue(var);
node.writeValue(createVariant());  // even better
```

chandryan:
> * Are the ref qualifiers in the getters really safe? If the data is borrowed, the move with rvalue ref qualifiers will move from the origin, right? Is that what you expect?

Ohh good point, that's unexpected and not safe at all 🙈 

> > I'm cautious with removing setter methods and forcing the user to always re-construct. It seems more fitting for scenarios such as lock guards and I would not rule out that they are needed with some code patterns. For example:
> > ```
> > Variant var;
> > 
> > if (conditionA)
> >   var.setValue(4);
> > else if (conditionB)
> >   var.setValue(4.0f);
> > else
> >   var.setValue("Hello");
> > 
> > node.writeValue(var);
> > ```
> 
> On the other hand re-contructing is the common way in the STL, e.g. with `std::variant`, `std::optional`... I would argue, that the shown example is not best practice anyway. I could be easily refactored with a factory function or even a immediately invoked lambda. As a result, the created `Variant` can be marked `const` or passed directly to another function as a rvalue reference:
> 
> ```c++
> const auto createVariant = [&] {
>   if (conditionA)
>     return Variant(4);
>   if (conditionB)
>     return Variant(4.0f);
>   return Variant("Hello");
> };
> 
> const auto var = createVariant();
> node.writeValue(var);
> node.writeValue(createVariant());  // even better
> ```

I agree it's not a good example and honestly I cannot think of a better one :D
Though the STL classes you mentioned do have a setter via the `emplace` function and I cannot think of a STL container class that doesn't allow modifying its underyling value(s) after construction.

chandryan:
> * Are the ref qualifiers in the getters really safe? If the data is borrowed, the move with rvalue ref qualifiers will move from the origin, right? Is that what you expect?

Ohh you're right, I would not expect that and it does not sound safe at all 🙈 

If I understood the C code correctly then ownership of this Variant container is defined at runtime, within the Variant. This is kind of odd from a C++ perspective, I think the question of ownership is usually already handled at the class level (e.g. `string` and `string_view`).

What would be a good way to handle this situation safely in C++? Maybe something like only allowing to return const refs in getters and moving is only possible by an additional function from the Variant API, which needs to check the underyling UA_Variant's storageType?

lukasberbuer:
> > * Are the ref qualifiers in the getters really safe? If the data is borrowed, the move with rvalue ref qualifiers will move from the origin, right? Is that what you expect?
> 
> Ohh you're right, I would not expect that and it does not sound safe at all 🙈
> 
> If I understood the C code correctly then ownership of this Variant container is defined at runtime, within the Variant. This is kind of odd from a C++ perspective, I think the question of ownership is usually already handled at the class level (e.g. `string` and `string_view`).
> 
> What would be a good way to handle this situation safely in C++? Maybe something like only allowing to return const refs in getters and moving is only possible by an additional function from the Variant API, which needs to check the underyling UA_Variant's storageType?

We could change the return type from `T&&` to `T` for the rvalue ref-qualified getters and make a runtime-check `encoding != UA_EXTENSIONOBJECT_DECODED_NODELETE` to decide if to move or not to move:
- if the `Variant` is a rvalue and owns the data, this will move the data and RVO will eliminate a second move.
- if the `Variant` is a rvalue and borrows the data, the data is copied.

I created a small demo in compiler explorer: https://godbolt.org/z/ej5Wqqo4h

lukasberbuer:
Just had an idea: Why not use the assignment operator instead of `setValue`? This would be the most elegant solution in my opinion?

chandryan:
> > > * Are the ref qualifiers in the getters really safe? If the data is borrowed, the move with rvalue ref qualifiers will move from the origin, right? Is that what you expect?
> > 
> > 
> > Ohh you're right, I would not expect that and it does not sound safe at all 🙈
> > If I understood the C code correctly then ownership of this Variant container is defined at runtime, within the Variant. This is kind of odd from a C++ perspective, I think the question of ownership is usually already handled at the class level (e.g. `string` and `string_view`).
> > What would be a good way to handle this situation safely in C++? Maybe something like only allowing to return const refs in getters and moving is only possible by an additional function from the Variant API, which needs to check the underyling UA_Variant's storageType?
> 
> We could change the return type from `T&&` to `T` for the rvalue ref-qualified getters and make a runtime-check `encoding != UA_EXTENSIONOBJECT_DECODED_NODELETE` to decide if to move or not to move:
> 
> * if the `Variant` is a rvalue and owns the data, this will move the data and RVO will eliminate a second move.
> * if the `Variant` is a rvalue and borrows the data, the data is copied.
> 
> I created a small demo in compiler explorer: https://godbolt.org/z/ej5Wqqo4h

I like that solution, it looks good :-)

include/open62541pp/types.hpp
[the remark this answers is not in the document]
chandryan: Outcome of discussion: Let's use `ReferenceTag`

include/open62541pp/types.hpp
```diff
@@ -888,59 +961,99 @@ class Variant : public TypeWrapper<UA_Variant, UA_TYPES_VARIANT> {
 public:
     using TypeWrapper::TypeWrapper;  // inherit constructors
 
+    // Universal constructors with default copy policy
+
+    // TODO:
+    // Allow implicit constructor?
+    template <typename T, typename X = std::enable_if_t<!std::is_same_v<T, Variant>, void>>
+    explicit Variant(T&& value) {
+        setValueCopy(std::forward<T>(value));
```
chandryan, line 970: If you look at the new universal constructors and the `setXyz` API side-by-side, they don't seem to match very well:

- The constructors copy by default, and you opt into referencing the data by using a reference tag
- The `setXyz` API references the data, and it feels like you opt into copying by using `setXyzCopy`

It feels confusing to have this kind of mismatch in the API. We could adapt `setValue` in a way that resembles the new constructors more closely: We could either turn around the API names to `setValue / setValueReference` or use a single `setValue` API with tag dispatch in the same way we do in the constructor.

From API perspective I find using a single `setValue` API with the same tag dispatch mechanism the most appealing, though that does not match well with the scalar and array get/set functions 😅 

What do you think? @lukasberbuer

lukasberbuer: To be honest, it doesn't bother me that much having a `ReferenceTag` and the suffix `Copy` for setters/getters, but I get your point.
On the one hand, I would like to stick to `set*Copy` and `get*Copy` to be close to the open62541 API. On the other hand, copy should be the default (safe) choice without any constructor tag.

Implementing a `setValue` with a tag and forwarding it either to `setScalar(Copy)` or `setArray(Copy)` should be easy to implement and could be done outside of this library.
My biggest concern is that users who are not so familiar with templates will be put off by such API choices.

include/open62541pp/types.hpp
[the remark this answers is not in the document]
chandryan: Now only `ReferenceTag` and an instance of it `reference` is left.

include/open62541pp/types.hpp
chandryan: Removed `ReferenceIfPossible`

include/open62541pp/types.hpp
chandryan: Changed API to `setValue` / `setValueCopy`

lukasberbuer:
Hi @chandryan,
In course of #459, I'm working on the builtin types, but skipped the `Variant` so far.
So far, we discussed changes in the variant policy and adding the unified API (constructor, setters and getters).

Shall we split this up into two PRs?
1. Simplify the `Variant` policy (copy as default, use tag for nocopy/reference) @lukasberbuer
2. Unified API by @chandryan 

If you agree, I would start working on the first PR.

chandryan:
> Hi @chandryan, In course of #459, I'm working on the builtin types, but skipped the `Variant` so far. So far, we discussed changes in the variant policy and adding the unified API (constructor, setters and getters).
> 
> Shall we split this up into two PRs?
> 
> 1. Simplify the `Variant` policy (copy as default, use tag for nocopy/reference) @lukasberbuer
> 2. Unified API by @chandryan
> 
> If you agree, I would start working on the first PR.

I had one issue left I could now resolve, I can finalize the preparations for this pull request in the next hours :) If you prefer to split it anway though, that's also okay with me 👍 

One point in the context of the `Variant` class I could not understand from the referenced issues: What would you like to do with methods, that have getters and setters:
- Remove only the get prefix of the getter and keep it for the setter or
- Keep prefixes for both getter and setter?

lukasberbuer:
> I had one issue left I could now resolve, I can finalize the preparations for this pull request in the next hours :) If you prefer to split it anway though, that's also okay with me 👍
> 
> One point in the context of the `Variant` class I could not understand from the referenced issues: What would you like to do with methods, that have getters and setters:
> 
> * Remove only the get prefix of the getter and keep it for the setter or
> * Keep prefixes for both getter and setter?

Great, I'm happy if you continue with the PR and looking forward to get it merge 😁

Let's update the getter names in another PR:
- setters have a `set` prefix
- getters don't have any prefix

I will add some more context to #459 😊

include/open62541pp/types.hpp
chandryan: I'm not sure if we had discussed your idea [using an assignment operator](https://github.com/open62541pp/open62541pp/pull/384#issuecomment-2434568042). I suppose it's probably not a good option if you prefer to go with `setValue `/ `setValueCopy`, and it would also require a reference tag?

chandryan: I'm not sure if we had discussed [your idea about using assignment operators](https://github.com/open62541pp/open62541pp/pull/384#issuecomment-2434568042). I suppose it's not a good option if you want to go with `setValue` / `setValueCopy`, and the assignment operators would also require a reference tag?

include/open62541pp/types.hpp
```diff
@@ -871,15 +871,23 @@ enum class VariantPolicy {
...
+static constexpr bool isCopyableOrConvertible = detail::isRegisteredType<T> ||
+    detail::isConvertibleType<T>;
+
 }  // namespace detail
 
+/// Tag used for constructing a Variant that references a value.
+struct ReferenceTag {};
+
+/// Tag instance used for constructing a Variant that references a value.
+static constexpr ReferenceTag reference{};
```
lukasberbuer, line 889: Should be `inline`

include/open62541pp/types.hpp
```diff
@@ -871,15 +881,23 @@ enum class VariantPolicy {
...
-    ReferenceIfPossible,  ///< Favor referencing but fall back to copying if necessary.
     // clang-format on
 };
 
 namespace detail {
-template <VariantPolicy>
-struct VariantHandler;
+
+template <typename T>
+static constexpr bool isCopyableOrConvertible = detail::isRegisteredType<T> ||
```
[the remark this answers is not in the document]
lukasberbuer: Yes :)

include/open62541pp/types.hpp
chandryan: Changed to `inline constexpr`

include/open62541pp/types.hpp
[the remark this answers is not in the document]
chandryan: Renamed this function to `isRegisteredOrConvertible` and an associated function to `assertIsRegisteredOrConvertible`

lukasberbuer:
> > > > * Are the ref qualifiers in the getters really safe? If the data is borrowed, the move with rvalue ref qualifiers will move from the origin, right? Is that what you expect?
> > > 
> > > 
> > > Ohh you're right, I would not expect that and it does not sound safe at all 🙈
> > > If I understood the C code correctly then ownership of this Variant container is defined at runtime, within the Variant. This is kind of odd from a C++ perspective, I think the question of ownership is usually already handled at the class level (e.g. `string` and `string_view`).
> > > What would be a good way to handle this situation safely in C++? Maybe something like only allowing to return const refs in getters and moving is only possible by an additional function from the Variant API, which needs to check the underyling UA_Variant's storageType?
> > 
> > 
> > We could change the return type from `T&&` to `T` for the rvalue ref-qualified getters and make a runtime-check `storage != UA_VARIANT_DATA_NODELETE` to decide if to move or not to move:
> > 
> > * if the `Variant` is a rvalue and owns the data, this will move the data and RVO will eliminate a second move.
> > * if the `Variant` is a rvalue and borrows the data, the data is copied.
> > 
> > I created a small demo in compiler explorer: https://godbolt.org/z/ej5Wqqo4h
> 
> I like that solution, it looks good :-)

Finally applied in #615 😊

PR #448, opened 2024-11-16
fix: zero initialization of unions or structs with unions

lukasberbuer:
The `Wrapper` template base class uses braced initialization, but braced initialization will only zero-initialize the first member of a union. If the first member of a union is small than the other union members, than the rest of the union remains uninitialized. This caused errors detected in #447.

Solution: Use `std::memset` for unions or structs that may contain unions.


lukasberbuer:
Sadly, there is no generic solution that maintains the `constexpr` default constructor (`memset` is not `constexpr`).
Currently, `NodeId` is the only type, that faces the issue. The zero-initialization of non-first members is fixed in #447.

PR #462, opened 2024-11-26
feat!: add node id parameter to DataSource callback functions

OgreTransporter:
I have the problem that I create nodes on my server in a loop, depending on which devices are currently connected. This means that my variables are not hard-coded. As a result, I cannot install any fixed callback functions for reading and writing the values. I therefore need the node ID to be able to assign the values.


include/open62541pp/plugin/nodestore.hpp
```diff
@@ -53,25 +53,27 @@ struct ValueBackendDataSource {
...
+     * @param id The ID of the node whose value was requested
      * @param value The DataValue that is returned to the reader
      * @param range If not empty, then the data source shall return only a selection of the
      *              (nonscalar) data.
      *              Set `UA_STATUSCODE_BADINDEXRANGEINVALID` in `value` if this does not apply
      * @param timestamp Set the source timestamp of `value` if `true`
      * @return StatusCode
      */
-    std::function<StatusCode(DataValue& value, const NumericRange& range, bool timestamp)> read;
+    std::function<StatusCode(NodeId id, DataValue& value, const NumericRange& range, bool timestamp)> read;
```
lukasberbuer, line 64: Please pass by `const NodeId&` to avoid copies.

include/open62541pp/plugin/nodestore.hpp
```diff
@@ -53,25 +53,27 @@ struct ValueBackendDataSource {
...
      * This function can be empty if the operation is unsupported.
      *
+     * @param id The ID of the node whose value is to be written
      * @param value The DataValue that has been written by the writer
      * @param range If not empty, then only this selection of (non-scalar) data should be written
      *              into the data source
      * @return StatusCode
      */
-    std::function<StatusCode(const DataValue& value, const NumericRange& range)> write;
+    std::function<StatusCode(NodeId id, const DataValue& value, const NumericRange& range)> write;
```
lukasberbuer, line 76: Pass by const ref.

src/server.cpp
```diff
@@ -387,9 +387,7 @@ static UA_StatusCode valueSourceRead(
     assert(nodeContext != nullptr && value != nullptr);
     auto& callback = static_cast<detail::NodeContext*>(nodeContext)->dataSource.read;
     if (callback) {
-        auto result = detail::tryInvoke(
-            callback, asWrapper<DataValue>(*value), asRange(range), includeSourceTimestamp
-        );
+        auto result = detail::tryInvoke(callback, NodeId(*nodeId), asWrapper<DataValue>(*value), asRange(range), includeSourceTimestamp);
```
lukasberbuer, line 390: Using the `NodeId` constructor will make a copy. Casting the native type to its wrapper type with `asWrapper<NodeId>(*nodeId)` avoids the copy.
Please add `nodeId != nullptr` to the assert statement above, so we don't accidentally dereference a `nullptr`.

include/open62541pp/plugin/nodestore.hpp
OgreTransporter: Fixed

include/open62541pp/plugin/nodestore.hpp
OgreTransporter: Fixed

src/server.cpp
OgreTransporter: Fixed

lukasberbuer:
Thanks for the updates 👍
There are still some compile errors/warnings in the CI pipeline to be fixed. Please let me know if you need support.

lukasberbuer:
@OgreTransporter did you try this PR? Can it be merged?

lukasberbuer (approved):

Issue #484, opened 2024-12-04
Where to specify the "ApplicationURI"?


Mebus:
Hallo, 

I am generating a certificate like this:

```
  char CLIENT_URI[] = "urn:mymachine:application";

  opcua::crypto::CreateCertificateResult client_cert = opcua::crypto::createCertificate(
      {
          opcua::String{"C=DE"},
          opcua::String{"O=SB"},
          opcua::String{"CN=client@mymachine"},
      },
      {
          opcua::String{"DNS:mymachine"},
          opcua::String{std::string("URI:").append(CLIENT_URI)},
      },
      opcua::crypto::CertificateFormat::DER);
```

However this results in this error:

```
The configured ApplicationURI does not match the URI specified in the certificate for the SecurityPolicy http://opcfoundation.org/UA/SecurityPolicy#Basic256Sha256
```

Where and how can I specify the ApplicationURI so that both match?

Mebus

lukasberbuer:
Hi @Mebus,
did you set the ApplicationURI of the `Client`? The functionality is currently not exposed in the C++ interface, but there is a simple workaround:
```cpp
opcua::Client client;

// with config() function since v0.16
asWrapper<opcua::ApplicationUri>(client.config().handle()->applicationUri) = opcua::String("...");

// with UA_Client_getConfig before v0.16
auto* config = UA_Client_getConfig(client.handle());
asWrapper<opcua::ApplicationUri>(config->applicationUri) = opcua::String("...");
```

Mebus:
Thanks! That seems to have changed something :-)

lukasberbuer:
Does it work?
PR #485 adds the missing `ClientConfig::setApplicationUri` function.

lukasberbuer:
Ups, I think it's the wrong setting:

```c
typedef struct {
    /* The description must be internally consistent.
     * - The ApplicationUri set in the ApplicationDescription must match the
     *   URI set in the certificate */
    UA_ApplicationDescription clientDescription;

    // ...

    /**
     * If the EndpointDescription has not been defined, the ApplicationURI
     * constrains the servers considered in the FindServers service and the
     * Endpoints considered in the GetEndpoints service.
     *
     * If empty the applicationURI is not used to filter.
     */
    UA_String applicationUri;

    // ...
} UA_ClientConfig;
```

So we need to change the `clientDescription.applicationUri`:
```cpp
auto* config = UA_Client_getConfig(client.handle());
[...]

PR #485, opened 2024-12-04
feat: ClientConfig::setApplicationUri

lukasberbuer:
Add missing function to set client's ApplicationURI.
See #484.

### DIFF
--- a/include/open62541pp/client.hpp
+++ b/include/open62541pp/client.hpp
@@ -94,6 +94,9 @@ public:
 
     /// Set message security mode.
     void setSecurityMode(MessageSecurityMode mode) noexcept;
+
+    /// Set the application URI, e.g. `urn:open62541.client.application`.
+    void setApplicationUri(std::string_view uri);
 };
 
 /* ------------------------------------------- Client ------------------------------------------- */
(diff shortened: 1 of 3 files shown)


lukasberbuer:
I intended to wrap the `UA_ClientConfig.clientDescription.applicationUri` instead of `UA_ClientConfig.applicationUri`.
`UA_ClientConfig.applicationUri` is used to filter servers by application URIs. I assume that this is a rarely used option.

```c
typedef struct {
    /* The description must be internally consistent.
     * - The ApplicationUri set in the ApplicationDescription must match the
     *   URI set in the certificate */
    UA_ApplicationDescription clientDescription;

    // ...

    /**
     * If the EndpointDescription has not been defined, the ApplicationURI
     * constrains the servers considered in the FindServers service and the
     * Endpoints considered in the GetEndpoints service.
     *
     * If empty the applicationURI is not used to filter.
     */
    UA_String applicationUri;

    // ...
} UA_ClientConfig;
```

sergioherrera7:
Currently I do something like:


(...)
    // Initialise client configuration
    UA_ClientConfig* client_config = new UA_ClientConfig();
    client_config->clientDescription.applicationUri = UA_String_fromChars(application_URI.c_str());
    // Initialize the opcua::Client instance using configuration
    opcua::ClientConfig clientConfig(std::move(*client_config));
    client_ = std::make_unique<opcua::Client>(std::move(clientConfig));
(...)


this feature would be nice to have.

Issue #509, opened 2024-12-14
Question about nullptr input in the Variant's "no copy" APIs


chandryan:
Hi @lukasberbuer,

the new Variant interface looks really slick, great work!! 😎 

I have a minor remark about the `nullptr` edge case that is currently used in the `Variant`'s `assign` functions as well as in the `ExtensionObject` constructor:

In the `ExtensionObject`'s constructor I think it makes sense to early-return with an effectively default-constructed object, though in the context of `Variant`'s assignment I find this behavior contra-intuitive. My expectation would be that a nullptr assignment resets the Variant, so that it's empty again. Similar to e.g. resetting a `std::function myFunction` via `myFunction = nullptr;`.

What do you think?

lukasberbuer:
Thank you @chandryan. It's all based on your work 😊
I totally agree, passing a `nullptr` to `Variant::assign` should clear the variant.

Do you want to provide a PR?

PR #510, opened 2024-12-15
feat: Change Variant nullptr assignment to clear avariant

chandryan:
This Pull Request addresses the nullptr assignment discussed in https://github.com/open62541pp/open62541pp/issues/509.

Contents:

- Changed the behavior of a nullptr assignment for Variants from no-op to clearing the variant.
- Additionally provided overloads for an assignment of `std::nullptr_t`.


lukasberbuer:
Thanks @chandryan.
I wonder if the `std::nullptr_t` overloads are a bit overkill. Do you know any STL type, that overloads assignments for `std::nullptr_t`?
I clearing the `Variant` is a use case, maybe we should expose the `clear` member function instead?

chandryan:
> Thanks @chandryan. I wonder if the `std::nullptr_t` overloads are a bit overkill. Do you know any STL type, that overloads assignments for `std::nullptr_t`? I clearing the `Variant` is a use case, maybe we should expose the `clear` member function instead?

I know at least about `std::function`: There are dedicated overloads available for nullptr assignments: https://en.cppreference.com/w/cpp/utility/functional/function/function

I agree with you that it's kind of overkill and the whole `nullptr` situation in general feels more like an undesired side effect of using pointer syntax for no-copy assignments :-(
On the other hand assigning a nullptr directly leads to a not well understandable error message, because the compiler tried the `const T&` overload. So adding a `std::nullptr_t` overload at least felt more consistent with the API.

I'm also fine with removing it again, whichever way you prefer.

lukasberbuer:
The overload resolution with `nullptr` is confusiong, I agree.
What about just deleting the overloads for `std::nullptr_t`?

```cpp
void assign(std::nullptr_t ptr) noexcept = delete;
void assign(std::nullptr_t ptr, const UA_DataType& type) noexcept = delete;
```

chandryan:
> The overload resolution with `nullptr` is confusiong, I agree. What about just deleting the overloads for `std::nullptr_t`?
> 
> ```c++
> void assign(std::nullptr_t ptr) noexcept = delete;
> void assign(std::nullptr_t ptr, const UA_DataType& type) noexcept = delete;
> ```

Great idea, changed as suggested :)

PR #523, opened 2025-01-03
fix!: Avoid redefinition of NOMINMAX macro for Windows systems

GiBg1aN:
If open62541pp is imported in a project which relies on such a macro, and it is not the first header to be imported, STL code may break.

If `NOMINMAX` is defined globally, then redefinition warning may be raised.
For instance with Visual Studio 2022 with MSVC compiler the following warning appears:
`warning C4005: 'NOMINMAX': macro redefinition`


lukasberbuer:
Great, thank you @GiBg1aN 👍

PR #590, opened 2025-04-24
feat: async operations to offload method calls to worker thread

lukasberbuer:
Method calls are executed in the event loop of the server. For the duration of the method execution, the event loop and server are blocked as described in #586. As a solution, method call execution can be outsourced to a worker thread. This functionality is called *async operations* in open62541. Two steps are required:

1. Enable async operations for an method node with:
    ```cpp
    useAsyncOperation(server, id, true /* enabled */)
    ```
2. Execute those async operations in worker thread(s) or even a thread pool, e.g.:
    ```cpp
    Server server;
    // ...
    std::thread worker([&] {
        while (true) {
            const std::optional<AsyncOperation> operation = opcua::getAsyncOperation(server);
            if (operation.has_value()) {
                opcua::runAsyncOperation(server, operation.value());
            } else {
                std::this_thread::sleep_for(std::chrono::milliseconds{10});
            }
        }
    });
    ```

This feature requires open62541 >= v1.1 and `UA_MULTITHREADING >= 100`.


src/server.cpp
```diff
@@ -513,4 +513,45 @@ UA_Server* getHandle(Server& server) noexcept {
...
+static void setAsyncOperationResult(Server& server, const T& response, void* context) noexcept {
+    static_assert(sizeof(T) <= sizeof(UA_AsyncOperationResponse));
+    UA_Server_setAsyncOperationResult(
+        server.handle(),
+        reinterpret_cast<const UA_AsyncOperationResponse*>(&response),  // NOLINT
+        context
+    );
+}
+
+bool runAsyncOperation(Server& server) {
```
17steen, line 533: This API limits the user to running the async operations on one thread, so it would be async in the sense that it doesn't run on the main thread, but all async methods would still happen synchronously in a way.

A user might want to poll for async requests, and submit them to a thread pool for example

17steen: In order to allow for the thread pool use case, getting the request, and executing it should be split.

lukasberbuer: You could run `runAsyncOperations` in multiple threads but this will keep all threads busy polling async operations.

What do you think about wrapping those async operations as an `std::packaged_task<void()>` and add a method `getAsyncOperation` that returns an optional task?
Any other ideas?

17steen: Busy polling would mean the threads are always in use and can't be allocated for other work, which means the workers in the work pool couldn't do any other work


the packaged_task option sounds like it would do the job

lukasberbuer: Please have a look at the last commit. I guess we don't need the overhead of `packaged_task` with the `<future>` header. A simple `struct AsyncOperation` retrieved by `getAsyncOperation` and passed to `runAsyncOperation` should do the job. Please review again 😊

src/server.cpp
```diff
@@ -513,4 +513,45 @@ UA_Server* getHandle(Server& server) noexcept {
...
+        context
+    );
+}
+
+std::optional<AsyncOperation> getAsyncOperation(Server& server) noexcept {
+    AsyncOperation operation{};
+    const bool hasRequest = UA_Server_getAsyncOperationNonBlocking(
+        server.handle(), &operation.type, &operation.request, &operation.context, nullptr
+    );
+    if (!hasRequest || operation.request == nullptr) {
```
17steen, line 538: Checking if operation.request is null sounds redundant, and if it were null, wouldn't it mean something went wrong

tests/services_method.cpp
```diff
@@ -116,4 +123,58 @@ TEMPLATE_TEST_CASE("Method service set", "", Server, Client, Async<Client>) {
...
+        const auto result = future.get();
+        CHECK(result.statusCode().isGood());
+        CHECK(result.outputArguments().at(0).to<uint64_t>() != getThreadId());
+    }
+    
+    SECTION("Async operation") {
+        useAsyncOperation(setup.server, methodId, true);
+        CHECK_FALSE(getAsyncOperation(setup.server).has_value());
+        auto future = services::callAsync(setup.client, objectsId, methodId, {}, useFuture);
+        std::this_thread::sleep_for(std::chrono::milliseconds{100});
```
17steen, line 169: 100ms sounds like some time to add to the test suite, 
maybe instead the test could poll getAsyncOperation for 100ms? So that the happy case is very short

only relevant if the tests happen synchronously i guess

17steen (approved):

17steen (approved):

PR #597, opened 2025-04-28
feat: DataTypeBuilder::addField overloads with manual offsets

17steen:
Sometimes it's not possible to get the member pointer at compile time, and offsets may need to be used. This is especially true when using reflection to generate offsets, in which case there is no way to get a member pointer at compile time.

### DIFF
--- a/include/open62541pp/datatype.hpp
+++ b/include/open62541pp/datatype.hpp
@@ -511,7 +511,7 @@ template <typename TArray, typename TSize>
 auto& DataTypeBuilder<T, Tag, U>::addArrayFieldWithOffset(
     const std::string_view fieldName,
     size_t offsetSize,
-    size_t offsetArray,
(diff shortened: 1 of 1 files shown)


lukasberbuer:
Great addition, thanks @17steen!

I have two concerns:
- The functions without offset are called `addField` for both scalar and array fields, but the functions with manual offets are called `addFieldWithOffset` and `addArrayFieldWithOffset`. Maybe we can make them all overloads of `addField`?
- Why is `TSize` provided as a template parameter? Open62541 requires size types to be of type `size_t`.

lukasberbuer:
I just looked it up, open62541 assumes always `size_t`:
https://github.com/open62541/open62541/blob/v1.3.15/src/ua_types_encoding_binary.c#L1390-L1396

Passing another size type than `size_t` would result in UB. We should definitely `static_assert` the size of the size field type.

PR #612, opened 2025-05-15
fix: access control with v1.4

### LINKED ISSUE #610: examples/access_control/server_access_control.c: Doesn't work when using latest open62541 release (1.4.11)
Not sure how relevant since 1.4.11 is not the default version used for this library.
But I wanted to use it and see if some other issues were gone.

steps to reproduce: 

1. checkout 1.4.11 version of open62541, run `server_accesscontrol` example.
2. try to log in with an opc ua client (e.g. uaexpert)
3. try user/user, anonymous, admin/admin

Gets this error: 
```
Error 'BadIdentityTokenInvalid' was returned during ActivateSession
```

None of the logins  work, anonymous login also doesn't work.

Tested the C library example and i can login with peter/peter123.

lukasberbuer:
Fixes two issues:
- `Error 'BadIdentityTokenInvalid' was returned during ActivateSession` in `server_accesscontrol` example (closes #610)
- `BadIdentityTokenInvalid` in CI with open62541 >= v1.4.6

### DIFF
--- a/src/plugin/accesscontrol_default.cpp
+++ b/src/plugin/accesscontrol_default.cpp
@@ -12,8 +12,7 @@ constexpr std::string_view policyIdAnonymous = "open62541-anonymous-policy";
 constexpr std::string_view policyIdUsername = "open62541-username-policy";
 
 static constexpr bool startsWith(std::string_view str, std::string_view prefix) noexcept {
-    const auto common = std::min(str.size(), prefix.size());
-    return str.substr(0, common) == prefix.substr(0, common);
+    return str.size() >= prefix.size() && str.substr(0, prefix.size()) == prefix;
 }
 
 AccessControlDefault::AccessControlDefault(bool allowAnonymous, Span<const Login> logins)
--- a/src/plugin/accesscontrol_default.cpp
+++ b/src/plugin/accesscontrol_default.cpp
@@ -1,5 +1,6 @@
 #include "open62541pp/plugin/accesscontrol_default.hpp"
 
+#include <cmath>
 #include <string_view>
 #include <utility>  // move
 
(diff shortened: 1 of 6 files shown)


src/plugin/accesscontrol_default.cpp
```diff
@@ -10,6 +11,11 @@ namespace opcua {
 constexpr std::string_view policyIdAnonymous = "open62541-anonymous-policy";
 constexpr std::string_view policyIdUsername = "open62541-username-policy";
 
+static constexpr bool startsWith(std::string_view str, std::string_view prefix) noexcept {
```
17steen, line 14: This isn't *exactly* startsWith in the sense that `startsWith("hel", "hello") -> true,

not sure if there's a more fitting name for this though

tests/client.cpp
```diff
@@ -138,7 +141,7 @@ TEST_CASE("Client connectAsync/disconnectAsync") {
     ServerRunner serverRunner{server};
     Client client;
 
-    bool connected = false;
+    static bool connected = false;  // use static to extend lifetime
```
17steen, line 144: Not sure from looking at this why `connected`'s lifetime has to be extended, does it have to do with how/when values are read with this testing framework?

17steen (approved):

tests/client.cpp
lukasberbuer: I got a warning from ASan stack usage after scope.. the callbacks shouldn't be called after the destruction of the `Server`. Have to dive deeper to fully understand this - or maybe its a false positive.

src/plugin/accesscontrol_default.cpp
lukasberbuer: You are right.. this would be a proper `startsWith` implementation that also works for this case:

```cpp
static constexpr bool startsWith(std::string_view str, std::string_view prefix) noexcept {
    return str.size() >= prefix.size() && str.substr(0, prefix.size()) == prefix;
}
```

Agree?

17steen (approved):

PR #631, opened 2025-05-23
fix: UAPP_HAS_NODESETLOADER for v1.4

lukasberbuer:


lukasberbuer:
Good point, I will prepare a v0.20.0 release this week.

PR #689, opened 2026-02-10
Check if CMAKE_CXX_STANDARD is set and do not change it for the target

GMellar:


lukasberbuer (dismissed):

lukasberbuer:
https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#requiring-language-standards

After reading this, a library shouldn't set global `CMAKE_CXX_STANDARD*`  settings at all. Just removed it completely.

PR #697, opened 2026-03-02
feat(node): add hasChildren() convenience function

ozlb:
feat(node): add hasChildren() convenience function

Add `Node::hasChildren()` to check whether a node has forward hierarchical references.

The implementation performs a single browse call with maxReferences = 1 to avoid collecting all references and handling continuation points when only a boolean result is required.

This provides a lightweight alternative to `!browseChildren().empty()`.

### DIFF
--- a/include/open62541pp/node.hpp
+++ b/include/open62541pp/node.hpp
@@ -609,6 +609,26 @@ public:
         return nodes;
     }
 
+	/// Check if the node has children (only local nodes).
+    bool hasChildren(
+        const NodeId& referenceType = ReferenceTypeId::HierarchicalReferences,
+        Bitmask<NodeClass> nodeClassMask = NodeClass::Unspecified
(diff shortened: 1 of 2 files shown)


ozlb:
After thorough testing, it appears the solution does not work as expected in all scenarios and requires further investigation.

lukasberbuer:
I like the idea. What are the edge cases?

ozlb:
Actually, the real reason isn’t clear yet, and I need to investigate further. In my browse implementation, I was using:
`if (child.browseChildren().size() > 0) {` 
which worked perfectly but was a little slow. I replaced it with: 
`if (child.hasChildren())` and, to validate, I also added: 
`assert(child.hasChildren() == (child.browseChildren().size() > 0));`
Result: for some nodes, this triggers an assertion failure.

PR #700, opened 2026-03-06
fix: UA_DataTypeArray with cleanup=false freed stack-allocated (or externally-owned) memory

LaDeKlint:
Using open62541pp with a lot of compiled nodesets, I got recurring memory errors in teardown when running tests under ASan. I think I have found the issue.

In open62541 v1.4 `UA_DataTypeArray` gained a **cleanup** boolean field that signals whether the node and its types array are heap-allocated and therefore owned by the library (and should be freed), versus being statically/stack-allocated or externally owned (and must not be freed).

Two functions in `datatype.cpp` ignored the cleanup flag and unconditionally freed memory:

1. `detail::clear(UA_DataTypeArray&)`

2. `detail::deallocate(const UA_DataTypeArray*)`

I have written tests that fail under ASan if the fixes are not implemented.

I have updated both functions to gate the freeing calls on cleanup, in case open62541 v>1.4

### DIFF
--- a/src/datatype.cpp
+++ b/src/datatype.cpp
@@ -45,8 +45,8 @@ void clear(UA_DataTypeArray& native) noexcept {
             native.typesSize,
             [](auto& type) { clear(type); }
         );
+        deallocateArray(const_cast<UA_DataType*>(native.types));  // NOLINT
     }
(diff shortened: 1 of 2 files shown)


lukasberbuer:
Thank you for the fix. Looks good 👍

PR #704, opened 2026-04-20
fix: prevent custom datatype use-after-free and invalid free

dmitrybuggenhout-rv:
[open62541pp_bug_repro.zip](https://github.com/user-attachments/files/26893921/open62541pp_bug_repro.zip)

    This MR fixes two custom datatype teardown bugs in open62541pp.
  >
    1. A use-after-free during server cleanup: open62541pp frees customDataTypes before UA_Server_delete() has finished cleaning node values that still reference that metadata.
    2. An invalid free for zero-member custom datatypes (for example enums): open62541pp stores an empty members array using the empty-array sentinel, but open62541 cleanup expects either
     nullptr or a real allocation.
  >
    Attached are two minimal repro scripts, one for each bug.
  >
    The fixes are related because the first change delegates custom datatype cleanup back to open62541, and the second ensures the resulting cleanup path is safe for zero-member 
    datatypes.

### DIFF
--- a/src/datatype.cpp
+++ b/src/datatype.cpp
@@ -75,6 +75,14 @@ UA_DataTypeMember copy(const UA_DataTypeMember& src) {
 }
 
 static void copyMembers(const UA_DataTypeMember* members, size_t membersSize, UA_DataType& dst) {
+    // open62541's UA_cleanupDataTypeWithCustom calls UA_free(type->members) directly without
+    // stripping the empty-array sentinel. Use nullptr for empty to avoid crashing cleanup on
+    // types with no members (for example enums).
+    if (membersSize == 0) {
+        dst.members = nullptr;
+        dst.membersSize = 0;
+        return;
+    }
     dst.members = detail::allocateArray<UA_DataTypeMember>(membersSize);
     dst.membersSize = membersSize;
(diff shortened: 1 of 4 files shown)


lukasberbuer:
Thank you @dmitrybuggenhout-rv 👍
Could you add a test to reproduce the error and show that the fix is working?

dmitrybuggenhout-rv:
@lukasberbuer Done.
Removing the fix commit results in the tests failing.

dmitrybuggenhout-rv:
I see the pre-commit hook did change some spelling. I'm assuming that is intended

lukasberbuer:
We can ignore the codegen error for now, but compatibility with open62541 v1.0 seems to be broken.

dmitrybuggenhout-rv:
Seems like in versions prior to 1.4 `UA_ServerConfig_clean` does not free customDataTypes, resulting in leaks.

Could work with #if #else blocks, but would it be an idea to drop support for some older versions?

lukasberbuer:
We have to include some `#if`/`#else` blocks to keep compatibility.

lukasberbuer:
Perfect, looks clean!

PR #710, opened 2026-06-19
fix: dangling references when using useDeferred

### LINKED ISSUE #709: Async service call with deferred token causes dangling references
Given the following sample code:

```c++
auto client_config = opcua::ClientConfig();
auto client = opcua::Client(std::move(client_config));
client.connect("opc.tcp://localhost:12345");

auto node_id = opcua::NodeId::parse("ns=2;s=My.Node.Id");
auto write_op = opcua::services::writeValueAsync(client, node_id, opcua::Variant{false}, opcua::useDeferred);

auto write_future = std::move(write_op)(opcua::useFuture);  // memory error here

// ...setup another thread to run/run_iterate...
auto write_result = write_future.get();
```

The call to the deferred async operation causes dangling references, resulting in using invalid memory during the deferred operation initiation.

From what I can tell, this is caused by [`sendRequestAsync()`](https://github.com/open62541pp/open62541pp/blob/ab3cc158e6883cdea8e9760043cbf643eeb0999f/include/open62541pp/services/detail/client_service.hpp#L101), which captures its initiation arguments *by reference* in the lambda passed to `AsyncServiceAdapter::initiate()`. With `useDeferred`, the lambda isn't called immediately and is instead packaged into the object returned to the caller. However, the `WriteRequest` passed to `sendRequestAsync()` is a temporary, and by the time the lambda is invoked, the temporary has been destroyed.

It seems to me that both of these async function adapters should *not* capture any references (besides maybe `client`), and instead pass-through the arguments to `asyncInitiate()` to potentially be decay-copied by `useDeferred` if needed.

boimart1:
Fixes #709.

When calling an async function with `useDeferred`, the initiation object isn't called immediately, it's wrapped and returned to the caller. For `AsyncServiceAdapter::initiate()` and the functions that called it, some of the initiation lambda's captures were references to objects that only existed for the duration of the call to the async function. Those objects would be destroyed by the time the deferred operation was actually started.

The proposed solution is to use `asyncInitiate()`'s existing mechanism to pass-through trailing arguments, rather than capturing any references whatsoever. Out of an abundance of caution, I did the same with the `Client&` arguments, with `std::ref` to ensure that the reference remains when it's decay-copied. For `Client` specifically, it would be ok to explicitly capture it by reference.

~~I also made a small change to `Span`: when updating `createMonitoredItems...Async()`, I found that the two existing Span constructors prevents construction of a `Span<T>` (where T is non-const) to be constructed from a `vector<T>` rvalue. `std::span` defines construction from a container with `span(Container&&)`, so I changed this Span's constructors to behave similarly.~~

Note that I didn't add any new test cases. I wasn't too sure where they would fit in, or exactly which operations to add tests for. Is `auto op = ...Async(..., useDeferred); op(useFuture)` (for some or all of the modified operations) something you would like me to add tests for somewhere?

### DIFF
--- a/include/open62541pp/services/detail/client_service.hpp
+++ b/include/open62541pp/services/detail/client_service.hpp
@@ -81,18 +81,18 @@ struct AsyncServiceAdapter {
         );
 
         return asyncInitiate<Response>(
-            [](auto&& handler, Client& client, auto&& initiation, auto&&... args) {
-                auto& catcher = opcua::detail::getExceptionCatcher(client);
+            [](auto&& handler, Client& innerClient, auto&& innerInitiation, auto&&... innerArgs) {
+                auto& catcher = opcua::detail::getExceptionCatcher(innerClient);
                 try {
                     // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks), false positive?
                     auto callbackAndContext = makeCallbackAndContext(
                         catcher, std::forward<decltype(handler)>(handler)
                     );
                     std::invoke(
-                        std::forward<decltype(initiation)>(initiation),
+                        std::forward<decltype(innerInitiation)>(innerInitiation),
                         callbackAndContext.callback,
                         callbackAndContext.context.get(),
-                        std::forward<decltype(args)>(args)...
+                        std::forward<decltype(innerArgs)>(innerArgs)...
                     );
                     // initiation call might raise an exception
                     // transfer ownership to the callback afterwards
@@ -116,11 +116,11 @@ auto sendRequestAsync(Client& client, const Request& request, CompletionToken&& 
         client,
         [](UA_ClientAsyncServiceCallback callback,
            void* userdata,
-           Client& client,
-           const Request& request) {
+           Client& innerClient,
+           const Request& innerRequest) {
             throwIfBad(__UA_Client_AsyncService(
-                opcua::detail::getHandle(client),
-                &request,
+                opcua::detail::getHandle(innerClient),
+                &innerRequest,
                 &getDataType<Request>(),
                 callback,
                 &getDataType<Response>(),
--- a/include/open62541pp/services/detail/client_service.hpp
+++ b/include/open62541pp/services/detail/client_service.hpp
@@ -81,8 +81,8 @@ struct AsyncServiceAdapter {
         );
 
         return asyncInitiate<Response>(
-            [](auto&& handler, Client& innerClient, auto&& innerInitiation, auto&&... innerArgs) {
-                auto& catcher = opcua::detail::getExceptionCatcher(innerClient);
+            [&client](auto&& handler, auto&& innerInitiation, auto&&... innerArgs) {
+                auto& catcher = opcua::detail::getExceptionCatcher(client);
                 try {
                     // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks), false positive?
                     auto callbackAndContext = makeCallbackAndContext(
@@ -102,7 +102,6 @@ struct AsyncServiceAdapter {
                 }
             },
             std::forward<CompletionToken>(token),
-            std::ref(client),
             std::forward<Initiation>(initiation),
             std::forward<Args>(args)...
         );
@@ -114,12 +113,11 @@ template <typename Request, typename Response, typename CompletionToken>
 auto sendRequestAsync(Client& client, const Request& request, CompletionToken&& token) {
     return AsyncServiceAdapter<Response>::initiate(
         client,
-        [](UA_ClientAsyncServiceCallback callback,
-           void* userdata,
-           Client& innerClient,
-           const Request& innerRequest) {
+        [&client](
+            UA_ClientAsyncServiceCallback callback, void* userdata, const Request& innerRequest
+        ) {
             throwIfBad(__UA_Client_AsyncService(
-                opcua::detail::getHandle(innerClient),
+                opcua::detail::getHandle(client),
                 &innerRequest,
                 &getDataType<Request>(),
                 callback,
@@ -129,7 +127,6 @@ auto sendRequestAsync(Client& client, const Request& request, CompletionToken&& 
             ));
         },
         std::forward<CompletionToken>(token),
-        std::ref(client),
         request
     );
 }
--- a/include/open62541pp/services/detail/client_service.hpp
+++ b/include/open62541pp/services/detail/client_service.hpp
@@ -64,15 +64,24 @@ struct AsyncServiceAdapter {
      * Initiate open62541 async client operation with user-defined completion token.
      * @param client Instance of type Client
      * @param initiation Callable to initiate the async operation. Following signature is expected:
-     *                   `void(UA_ClientAsyncServiceCallback callback, void* userdata)`
+     *                   ```
+     *                   void(
+     *                       UA_ClientAsyncServiceCallback callback,
+     *                       void* userdata,
+     *                       Args... args)
+     *                   ```
      * @param token Completion token
      */
-    template <typename Initiation, typename CompletionToken>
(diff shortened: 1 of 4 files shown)


lukasberbuer:
Thank you for the fast fix. A few thoughts:

- **The fix can be simplified (?)**

  The trailing-args mechanism works, but it changes the `AsyncServiceAdapter::initiate` signature, adds variadic complexity to the inner lambda, requires `std::ref(connection)` at every call site, and forces the `Span` constructor change. All of this can be avoided.
  
  The minimal fix is to replace `[&]` with explicit captures: copy the local variables that would dangle, and keep `client`/`connection` captured by reference. Capturing a `Client&` by reference is always safe since the caller must keep the client alive for any async operation anyway:
  
  ```cpp
  // sendRequestAsync
  [&client, request](UA_ClientAsyncServiceCallback callback, void* userdata) {
      throwIfBad(__UA_Client_AsyncService(opcua::detail::getHandle(client), &request, ...));
  }
  ```
  
  The same pattern applies to the other lambdas. `AsyncServiceAdapter::initiate` stays a simple two-parameter function and `Span` needs no changes.
  
  Or do I miss something?

- **On the `Span` constructor change**

  With the original two constructors, `Span<T>` (non-const `T`) from a container rvalue was a compile error, an intentional guard against dangling mutable spans. Collapsing to a forwarding reference removes that guard. If the explicit-capture approach is used, this change is not needed at all.

- **On `std::ref(client)` in the current approach**

  As you noted, capturing `client` by reference in the initiation lambda is fine. Both approaches store a reference to the client in the deferred op, so they are semantically equivalent. The stateless lambda is cleaner, but not worth the added complexity at every call site.

boimart1:
On each point:

> * **The fix can be simplified (?)**
>   The trailing-args mechanism works, but it changes the `AsyncServiceAdapter::initiate` signature, adds variadic complexity to the inner lambda, requires `std::ref(connection)` at every call site, and forces the `Span` constructor change. All of this can be avoided.
>   The minimal fix is to replace `[&]` with explicit captures: copy the local variables that would dangle, and keep `client`/`connection` captured by reference. Capturing a `Client&` by reference is always safe since the caller must keep the client alive for any async operation anyway:
>   ```c++
>   // sendRequestAsync
>   [&client, request](UA_ClientAsyncServiceCallback callback, void* userdata) {
>       throwIfBad(__UA_Client_AsyncService(opcua::detail::getHandle(client), &request, ...));
>   }
>   ```
> 
>   The same pattern applies to the other lambdas. `AsyncServiceAdapter::initiate` stays a simple two-parameter function and `Span` needs no changes.
>   Or do I miss something?

This is mostly based on my experience using `boost::asio`, exploring its codebase, and creating my own async operations that integrate with the framework.

Behaviourally, the change you propose indeed leads to the same outcome. The difference is in the details: capturing by value incurs an extra copy of the request, so you'd rather want to move the local `request` into the lambda captures. Then if shadowing rules apply to `a = std::move(a)` captures, you still need a new name for the captured variable, say `innerRequest`. And in cases where the `open62541` function takes a non-const pointer, e.g. `UA_Client_MonitoredItems_createDataChanges_async()`, you'd need to apply `mutable` to the lambda to ensure the captured object is non-const. Overall you end up with:

```c++
// sendRequestAsync
[&client, innerRequest = std::move(request)](UA_ClientAsyncServiceCallback callback, void* userdata) { /*...*/ }

// createMonitoredItemsDataChangeAsync
[&connection,
 innerRequest = std::move(request),
 innerContextsPtr = std::move(contextsPtr),
 innerDataChangeCallbacks = std::move(dataChangeCallbacks)
 innerDeleteCallbacks = std::move(deleteCallbacks)](UA_ClientAsyncServiceCallback callback, void* userdata) mutable { /*...*/ }
```

Beyond that, there are some benefits to using the variadic arguments over captures.

- In the usual case where the initiation lambda is invoked within the context of the call to `asyncInitiate()`, objects are only perfect-forwarded, and are never actually move-constructed into the lambda. This is only a slight performance gain, but I think it's still worth noting.

- In the case where the initiation lambda is deferred, arguments that are passed-through to the call to `asyncInitiate()` are easier to make behave optimally than those that are captured within the initiating lambda. This is true with `boost::asio::deferred`, which I've adapted to be usable with `opcua`'s async operations. In short, `deferred` preserves expectations with regards to deferred async ops and their value categories. That is, given `auto op = opcua::___Async(..., deferred);`, then `op(token)` can potentially be performed multiple times if the async operation supports it, whereas `std::move(op)(token)` steals the op's inner state to perform the operation. This is done by `deferred` returning an object with different overloads for `operator()(Token) const&` and `operator()(Token) &&`. Using the trailing variadic arguments allows to piggyback on `deferred`'s implementation to automatically have the category-preserving behaviour for all such arguments. Whereas if you wanted to have the same category-preserving behaviour with captured arguments, the initiation function object would need to be a proper struct with the same two `operator() const&/&&` overloads, or a lambda with C++23's `this auto&& self` + `std::forward_like<decltype(self)>(capturedObj)`, increasing the overall complexity to provide the same guarantees. (Note that `opcua::useDeferred`'s current implementation doesn't preserve value categories, initiating it always steals the deferred op's state).

In general, I believe that in a generic library like this, some extra complexity in the library code is justified if it leads to more well-behaved code on the user side, no matter what (valid) combination of weird things the user throws at it.

(edit) But to be sure, I'm aware that there is little-to-no benefit to this approach compared to the lambda captures in these specific cases. It's just a matter of following the same conventions I'd use in my other Asio-code, and I'm not opposed to changing it here.

> * **On the `Span` constructor change**
>   With the original two constructors, `Span<T>` (non-const `T`) from a container rvalue was a compile error, an intentional guard against dangling mutable spans. Collapsing to a forwarding reference removes that guard. If the explicit-capture approach is used, this change is not needed at all.

The change to `Span` is somewhat orthogonal to the rest of the fix, I can revert it if you prefer. There are alternatives to using `Span` in the way I did, no matter if the lambda captures the callbacks or takes them as parameters.

~~Still, I'd argue that constructing a span from a temporary is desirable, regardless of if `T` is const or not. Allowing it permits more valid use-cases of `Span`, for example here, where the open62541 C API takes a "list" by non-const pointer, even if it doesn't modify the list. `Span` is a view type, so it already comes with the caveat that users must be careful to not cause dangling, and so I don't see much difference in the risk between constructing `Span<const T>` or `Span<T>` from a rvalue.~~

> * **On `std::ref(client)` in the current approach**
>   As you noted, capturing `client` by reference in the initiation lambda is fine. Both approaches store a reference to the client in the deferred op, so they are semantically equivalent. The stateless lambda is cleaner, but not worth the added complexity at every call site.

:+1: I'll change the `Client&` back to explicit ref captures.

I'll wait for further feedback before changing anything else.

boimart1:
On the `Span` change: I've just done my homework, it looks like `std::span` disallows the same conversion of `span<NonConstT>` from rvalue range (unless it's a borrowed range). I missed the concept requirement when I look at the cppreference page the first time 😅 I'll revert that change too.

boimart1:
Instead of changing the Span constructor, I've changed my Span usage to `Span<void* const>` and `Span<const UA_...Callback>`. This does mean that I end up with `const_cast<void**>` and `const_cast<UA_...Callback*>` at the C API callsite. I believe that this is ugly but correct, since the C API doesn't actually want to write to the pointers (and I think it's at least a bit better than the `reinterpret_cast` that was there before*). If one day the C API instead took `void* const*` and `UA_...Callback const*` for its "list of pointer" parameter types, thus properly "advertising" its intentions with these parameters, then the `const_cast`s would become unnecessary.

That said, I can also get rid of the `const_cast`s if I end up capturing the vectors in the lambdas and marking them mutable.

*Getting rid of the `reinterpret_cast` was thanks to changing the `vector<MonitoredItemContext*>` to `vector<void*>`, actually! Sorry if my changes are getting a bit out of hand 😄

lukasberbuer:
This looks fine to me, thank you for your efforts 😊

PR #717, opened 2026-07-11
Prevent UA_Server_run_iterate from being called concurrently with UA_Server_run_shutdown

karzhenkov:
Calling to `UA_Server_run_iterate` from `Server::runIterate` should be protected by `mutexRun`. Otherwise, these functions may be called concurrenly. Additionally, I suggest removing the checks of `context().running` that seem duplicated.


lukasberbuer:
Great fix, thanks!
The runIterate method now acquires a lock for each call, which can be expensive when building custom loops. Do you see any chance to avoid this?

karzhenkov:
I don't see any other easy way to do this reliably. Most likely, there is no cheaper option.

PR #719, opened 2026-08-26
fix: emit standard textual format from toString(NodeId)

### LINKED ISSUE #647: opcua::toString(NodeId) string with double-quotes
Hi all.

I try version open62541pp@0.19.0 vs open62541@v1.4.12.

I look that opcua::toString(NodeId)   place a double-quote in front and the end of the string

So i get "\"i=84\""  instead "i=84" because UA_print is used.

UA_NodeId_print should be the choice

Best regards
  MZ

Onwcan:
toString(NodeId) went through the generic UA_print, which wraps the identifier in quotes -- "i=84" rather than i=84. That output is not accepted by NodeId::parse, so the library's own print and parse functions could not round-trip: every NodeId type produced BadDecodingError when fed back in.

Use UA_NodeId_print and UA_ExpandedNodeId_print, which emit the format documented by NodeId::parse and by OPC UA Part 6, section 5.3.1.10.

The two-argument toString(object, type) overload is an explicit request for the generic representation and is unchanged.

Guarded on open62541 v1.1 (NodeId) and v1.2 (ExpandedNodeId), the versions that introduced the respective print functions; older versions keep the previous behaviour. Neither function requires UA_ENABLE_TYPEDESCRIPTION or UA_ENABLE_JSON_ENCODING, so toString now also works for these types in configurations where the generic overload returns an empty string.

Closes #647


lukasberbuer:
Thank you, great job 😊

Onwcan:
My pleasure! Glad I could help 😊

PR #721, opened 2026-08-28
fix: replace internal __UA_Server_addNode with public API, Closes #651

### LINKED ISSUE #651: Invalid call of an internal function
https://github.com/open62541pp/open62541pp/blob/3c6dd1f83463c14f737430b75b6bc4d01038dcb7/src/services_nodemanagement.cpp#L48

The internal function `__UA_Server_addNode` should not be used. It will be removed in the next version (see [commit 5a0b066](https://github.com/open62541/open62541/commit/5a0b0661353d210855b05cdacdaef3716c5aeb9f) in master branch). The current version already contains a note that it should not be used:

https://github.com/open62541/open62541/blob/v1.4.12/include/open62541/server.h#L1293

Onwcan:
## Problem

`services::addNode<Server>` called `__UA_Server_addNode`. That function is
internal to open62541 — its declaration is prefixed with *"Don't use this
function. There are typed versions as inline functions."* — and it is removed
from the public header in master (open62541/open62541@5a0b066), so open62541pp
would stop building against upcoming versions.

## Fix

Dispatch on the node class to the type-specific functions of the public API
(`UA_Server_addObjectNode`, `UA_Server_addVariableNode`, ...), which is what the
upstream header directs callers to do.

I deliberately did *not* use the generic `UA_Server_addNode_begin` /
`_finish` pair, even though it is public and takes the same arguments: it
acquires the server lock twice, leaving a window in which a half-constructed
node is observable. The typed functions keep the whole AddNodes operation under
a single lock, matching the previous behaviour. Their signatures are unchanged
from v1.0 through master, so the whole compatibility matrix is covered.

Since the typed functions take the attributes by value, the attributes are
decoded to the type matching the node class; a mismatch returns
`BadNodeAttributesInvalid` — the same status code the AddNodes service returns.

## Behaviour differences worth reviewing

1. `NodeClass::Unspecified` on a Server now returns `BadNodeClassInvalid`
   instead of `BadOutOfMemory`. The old code inherited `BadOutOfMemory` from
   open62541 failing at node allocation before it could report the real cause.
   The Client path still returns `BadOutOfMemory`, so the new test asserts the
   exact code only for Server.
2. `typeDefinition` is now ignored for `ObjectType`, `ReferenceType`,
   `DataType`, `View` and `Method`, because the typed upstream functions do not
   accept one. All of open62541pp's own wrappers for those node classes already
   pass an empty NodeId, so the public API is unaffected.
3. With `UA_ENABLE_METHODCALLS=OFF`, `NodeClass::Method` returns
   `BadNotSupported`, as `UA_Server_addMethodNode` does not exist in that
   configuration. Note that this configuration does not currently build at all
   for an unrelated reason (`UAPP_HAS_ASYNC_OPERATIONS` in `config.hpp.in` does
   not account for `UA_ENABLE_METHODCALLS`, so `src/server.cpp` fails to
   compile). Happy to open a separate PR for that.

## Testing

Existing tests already cover `addNode` for every node class across Server,
Client and Async<Client>; they pass unchanged. Added two sections covering the
new validation paths (invalid node class, attributes of a mismatching node
class).

Verified locally with GCC 15 / Ninja: full suite green in the default config and
with `-DUA_MULTITHREADING=100 -DUAPP_BUILD_EXAMPLES=ON`; all examples build;
clang-format clean; clang-tidy reports no new warnings.

Closes #651


lukasberbuer:
Very clean, thank you 🙏

Onwcan:
You're very welcome...

PR #722, opened 2026-08-28
fix: build with UA_ENABLE_METHODCALLS=OFF

Onwcan:
## Problem

Building with `-DUA_ENABLE_METHODCALLS=OFF` fails to compile:

    src/server.cpp:556:41: error: 'response' was not declared in this scope

`UAPP_HAS_ASYNC_OPERATIONS` in `config.hpp.in` only checks the open62541
version and `UA_MULTITHREADING`, so `runAsyncOperation()` is compiled even
without method calls. It declares a `CallMethodResult`, whose wrapper lives
inside `#ifdef UA_ENABLE_METHODCALLS` in `ua/types.hpp`, so the declaration
fails and the next line errors on the undeclared variable.

## Fix

Guard `UAPP_HAS_ASYNC_OPERATIONS` with `UA_ENABLE_METHODCALLS`, following the
same pattern as the neighbouring `UAPP_HAS_DATAACCESS` and
`UAPP_HAS_ASYNC_SUBSCRIPTIONS`. Async operations exist solely to service
method calls — `runAsyncOperation` needs `UA_Server_call`, the only operation
type is `UA_ASYNCOPERATIONTYPE_CALL`, and `useAsyncOperation` is already
documented as "Only supported for method nodes" — so the whole feature is
meaningless without them.

Note that open62541 itself guards the async-operation C API only by
`UA_MULTITHREADING`, so those types do remain declared; it is open62541pp's
own `CallMethodResult` wrapper that disappears.

The `Node` test fixture also created a method node unguarded, so that and the
two sections using it are now guarded. The method examples needed no change —
`examples/CMakeLists.txt` already wraps them in `if(UA_ENABLE_METHODCALLS)`.

## Scope: this fixes compilation, not the whole configuration

While verifying, I found two further blockers that live in open62541, not here:

1. With the default `UA_NAMESPACE_ZERO=REDUCED`, the generated namespace zero
   contains 7 method nodes that fail to instantiate, so `UA_Server_new()`
   returns `NULL` with *"Initialization of Namespace 0 failed with
   BadTargetNodeIdInvalid"*. I confirmed this with a plain C program linked
   only against open62541 — no open62541pp involved.
2. `UA_ENABLE_PUBSUB_INFORMATIONMODEL=ON` fails to link, as
   `src/pubsub/ua_pubsub_ns0.c` calls `setMethodNode_callback`, which is only
   compiled with method calls enabled.

With `UA_NAMESPACE_ZERO=MINIMAL` and the PubSub information model disabled, a
server starts fine (`UA_Server_run_startup -> Good`) — which is the kind of
constrained build where you would disable method calls in the first place.
So this PR makes open62541pp compile in a configuration that is genuinely
usable; it does not claim `METHODCALLS=OFF` works with the default namespace
zero. Mentioning it in case you want to add a CI job — it would need those two
extra options.

## Testing

- `METHODCALLS=OFF` + `NAMESPACE_ZERO=MINIMAL`: library and all 17 examples
  build `-Werror` clean; server starts.
- `METHODCALLS=OFF` + default `REDUCED` NS0: everything now compiles (it did
  not before); runtime blocked upstream as described.
- `METHODCALLS=ON` + `UA_MULTITHREADING=100`: 3853 assertions in 227 test
  cases pass, 24 examples build — the change is a no-op here.
- clang-format clean.


lukasberbuer:
Looks good, thanks 👌

Onwcan:
Thanks! Could you also take a look at #721 when you have a chance? It fixes #651. It touches the same method-call/node-management area, and #722 also removes the unrelated UA_ENABLE_METHODCALLS=OFF build blocker mentioned there.

PR #728, opened 2026-09-08
fix: format NumericRange independently of the global locale

vpfkfl753:
`NumericRange` formatting inherits the global C++ locale through `std::ostringstream`. With thousands grouping enabled, index `1000` becomes `1,000`. Parsing that string produces two dimensions (`1`, `0`) instead of the original index.

Use the classic locale for the range formatter. In the [OPC UA range syntax](https://reference.opcfoundation.org/Core/Part4/v105/docs/7.27), commas separate dimensions.

The regression test uses a custom `std::numpunct` facet, so it does not depend on installed locales. It covers a single index, a range, multiple dimensions, and parsing the generated string.

Tested on Linux with Clang 21 and libstdc++ 15: four assertions fail before the fix; all 226 CTest tests pass after it. clang-format passes.


vpfkfl753:
The `NumericRange` test passes in the [GCC 11 job](https://github.com/open62541pp/open62541pp/actions/runs/34220450973/job/102042075740), along with all 229 tests. Its gcovr report does not list the new line (102) as missed.

I also ran `NumericRange` locally with `UAPP_ENABLE_COVERAGE=ON` and GCC 15.2. gcov records seven executions of `ss.imbue(...)`. Both normal branches are covered; the two uncovered branches are marked `throw: true`. This matches Codecov's report of one partially covered line and zero missing lines.

